PR #48, opened 2017-03-02
docs(docs): Update docs for new manager UI and editor module

landonreed:
@mattwigway @demory @trevorgerhardt @evansiroky 

docs from PR can be viewed here: http://conveyal-data-tools.readthedocs.io/en/user-doc-update/


evansiroky:
When I run `mkdocs serve` from the folder `datatools-ui\docs` I get the following error:

> INFO    -  Building documentation... 
> ERROR   -  Config value: 'docs_dir'. Error: The 'docs_dir' should not be the parent directory of the config file. Use a child directory instead so that the config file is a sibling of the config file. 
>
> Aborted with 1 Configuration Errors!

evansiroky:
Nice.  I'd recommend adding an entry to your travis file to make sure the docs always get built without error.  I did so for commute: https://github.com/conveyal/commute/blob/docs/.travis.yml

landonreed:
Gotcha. Thanks @evansiroky.

PR #50, opened 2017-03-07
Fix lint errors

trevorgerhardt:


lib/gtfsplus/components/GtfsPlusVersionSummary.js
```diff
@@ -43,21 +43,33 @@ export default class GtfsPlusVersionSummary extends Component {
...
     const editingIsDisabled = !user.permissions.hasFeedPermission(feedSource.organizationId, feedSource.projectId, feedSource.id, 'edit-gtfs')
     const publishingIsDisabled = !user.permissions.hasFeedPermission(feedSource.organizationId, feedSource.projectId, feedSource.id, 'approve-gtfs')
     const header = (
-      <h3 onClick={() => {
-        if (!this.state.expanded) gtfsPlusDataRequested(version)
-        this.setState({ expanded: !this.state.expanded })
-      }}>
+      <h3
+        onClick={this._toggleExpanded}
+        onKeyDown={(e) => {
```
evansiroky: Seems like this keyDown handler is implemented the same way in a lot of places.  We should make a helper for the default action.

landonreed: Yea, maybe a higher order component for accessible interactions?

trevorgerhardt: I would say create a ticket and we should consider that in the future but leave it out for now.

PR #51, opened 2017-03-07
Dep bump

landonreed:
Bumped out-of-date dependencies and removed unused ones.

This includes Leaflet 0.7 -> 1.0.3, which included some work on ControlPoint to make sure dragging works OK.

There was also some work to make sure the GridLayers are working properly.

Additionally (sorry for the bloat):

- removal of a few unused files (ManagerNavbar, DatatoolsNavbar)
- fix for MTC to request feeds after third party sync


evansiroky:
let's merge the lint fixes into this branch and then review

lib/editor/actions/map/index.js
```diff
@@ -23,6 +23,50 @@ export function addControlPoint (controlPoint, index) {
   }
 }
 
+export function dragControlPointEnded (index, controlPoint, evt, pattern) {
+  return {
+    type: 'HANDLE_CONTROL_POINT_DRAG_END',
```
evansiroky: Is there a reducer that handles this action?

landonreed: No, it's all really handled below in `handleControlPointDragEnd`.

lib/manager/components/NotesViewerPanel.js
```diff
@@ -33,16 +32,29 @@ export default class NotesViewerPanel extends Component {
...
-      }}>
+      <h3
+        onClick={this._toggleExpanded}
+        onKeyDown={(e) => {
+          if (document.activeElement === e.target && e.which === 13) {
+            this._toggleExpanded()
+          }
+        }}
+        role='button'
+        tabIndex={0}
```
evansiroky: We should come up with a convention about what to use for clickable items.  Either an `a` or `button` within this h3 would be more appropriate and then we wouldn't need the onKeyDown handler.

lib/editor/actions/map/index.js
evansiroky: Let's get rid of these unused actions.

lib/manager/components/NotesViewerPanel.js
landonreed: Oh boy, NotesViewerPanel is old code... isn't even imported anywhere.  Deleting.

PR #57, opened 2017-09-12
Fix alert desc f and r

landonreed:
When first mapping alerts received from the MTC database to the UI data format, a find and replace operation is performed on the alert description field to replace extra new line chars.  This was failing for an alert that had a null description.  It's still unclear where this alert came from or why the field was null (generated by some other external process?).


landonreed:
Wrong repo.

PR #391, opened 2019-01-07
Feed version tabs and home fixes

evansiroky:
- Make it so there is no error when loading patterns in the feed version navigator.
- Make a project load after the home page mounts


evansiroky:
I can't approve my own PR, but your changes look good.

Issue #396, opened 2019-01-22
After authentication, app redirects to home rather than previous URL


landonreed:
It appears that during flow refactoring the string variable `redirectOnSuccess` was changed to a boolean:

https://github.com/conveyal/datatools-ui/blob/a6cb68607877120d4c4a1f753355ccaa1026ac3d/lib/common/containers/App.js#L68-L70

https://github.com/conveyal/datatools-ui/commit/a755e4546a1196b2c585c4f2e20faacbe3d5d77a#diff-12752f290ea607f91209066fb17cb621

Issue #395, opened 2019-01-17
Container refactor issues: Editor session lock not releasing and pattern shape disappearing


landonreed:
There appears to be something wrong with the editor session lock and pattern geometries (perhaps following the container refactor).

1. If I open two tabs for the same feed in the editor, and then attempt to "re-lock" the feed in the second tab, it doesn't seem to emit any action.
2. For some trip patterns, selecting the pattern in the editor will cause the shape to disappear. There seems to be an issue with the selector that handles the pattern's shape points and control points.

PR #401, opened 2019-01-24
Container refactor fixes + pattern editor refresh

### LINKED ISSUE #397: Pattern Editor: Removing non-stop control point throws error
Clicking a control point to remove it causes an error to be thrown. Upon further inspection, it looks like we're not handling the recalculated shape segments correctly. 

https://github.com/conveyal/datatools-ui/blob/a6cb68607877120d4c4a1f753355ccaa1026ac3d/lib/editor/util/map.js#L495-L504

`newSegment.coordinates` should not be spread in the above statement. Rather, it needs to be simply spliced in as an array of coordinates because the correct type of pattern coordinates is `Array<Coordinates>` (e.g., `[[[1, 0], [0, 1]], [[1, 1], [0, 0]]]`.

landonreed:
This PR addresses a few items discovered while working on some fixes for Pace Bus.

- fixes a few issues encountered with pattern editing (#395, #400, #399, fixes #397)
- fixes the editor lock overwrite functionality (also #395) and the redirect on login (#396)
- tweaks the layout of the pattern shape editor so that it's a bit more intuitive (see screengrab)
- replaces the direction icons for the pattern geometry with `leaflet-textpath` for better performance/legibility (see second screengrab)
- changes some components to `PureComponents` to prevent them from rendering as frequently (I'm not sure if this is effective in all the cases, so we may need to revisit some of the components, e.g. Sidebar).

image

image


evansiroky:
The CI runs are failing due to unresolved flow type errors.  Please fix those.

lib/editor/selectors/index.js
```diff
@@ -125,6 +128,20 @@ const getPresentControlPoints = (state: AppState) => {
   return state.editor.editSettings.present.controlPoints
 }
 export const getTripCounts = (state: AppState) => state.editor.data.tables.trip_counts
+export const getActivePatternStops = createSelector(
```
evansiroky: Please write a unit test for this selector.

lib/editor/selectors/index.js
```diff
@@ -360,10 +381,10 @@ function getControlPointsFromShapePoints (shapePoints: Array<ShapePoint>) {
 function addPatternStopsToShapePoints (
   oldShapePoints: Array<ShapePoint>,
   patternStops: Array<PatternStop>,
-  stops: Array<any>, // TODO: use more exact type
+  activePatternStops: Array<GtfsStop>,
```
evansiroky: Would be fantastic to have a unit test for this function too.  Of course it'd be awesome to have unit tests for each possible scenario, but even a test that sent the oldShapePoints variable with a length being less than 2 is a good starting point as this big function has 0% coverage.

lib/editor/util/gtfs.js
```diff
@@ -165,7 +165,7 @@ export const generateProps = (component: string, editorState: any): any => {
   }
 }
 
-export function getEntityBounds (entity: any, offset: number = 0.005): ?latLngBounds {
+export function getEntityBounds (entity: any, stops?: Array<GtfsStop>, offset: number = 0.005): ?latLngBounds {
```
evansiroky: This could also use a very basic unit test.

lib/editor/util/map.js
```diff
@@ -288,7 +288,7 @@ export async function recalculateShape ({
   index: number,
   newPoint?: LatLng,
   pattern: Pattern,
-  patternCoordinates: Coordinates,
+  patternCoordinates: Array<Coordinates>,
```
evansiroky: Another great candidate for adding tests.  Especially since there is already a file `lib/editor/util/__tests__/map.js_hold` with test cases for this exact function.

evansiroky:
The "hide inactive segments" checkbox could be greyed out unless there are some currently active segments.  Otherwise it's a little confusing since it's hiding everything all segments.

landonreed:
https://github.com/conveyal/datatools-ui/pull/401#issuecomment-458434242 is resolved with the [latest version of gtfs-lib](https://github.com/conveyal/gtfs-lib/releases/tag/v4.2.5).

landonreed:
@evansiroky, I have finally addressed your comments here, but this PR doesn't add unit tests. I think it's a higher priority to get this merged in though at this point to avoid any more dust gathering.

evansiroky:
https://github.com/conveyal/datatools-ui/pull/401#issuecomment-458434242 does not appear to be fixed while running datatools-server with gtfs-lib 4.2.5. Please see video posted in company slack.

lib/editor/util/__tests__/map.js
```diff
@@ -56,23 +63,24 @@ describe('editor > util > map >', () => {
       makeTestCase({
         editType: 'update',
         index: 0,
-        mapzenResponeFile: './fixtures/mapzen-response-update-first-stop.json',
+        // https://
```
evansiroky: please remove comment

lib/editor/util/__tests__/__snapshots__/map.js.snap
```diff
@@ -0,0 +1,1244 @@
...
+      ],
+      Array [
+        -77.922112,
+        42.088844,
+      ],
+      Array [
+        -77.922135,
+        42.089099,
+      ],
+    ],
```
evansiroky: The coordinates in this array don't go all the way to the new last stop. Is this the desired behavior?

This is what the coordinates look like before deleting the last stop:

Screen Shot 2019-04-02 at 5 28 57 PM

And here are those coordinates after:

Screen Shot 2019-04-02 at 5 29 33 PM

lib/editor/util/__tests__/__snapshots__/map.js.snap
```diff
@@ -0,0 +1,1244 @@
...
+        -77.922974,
+        42.095657,
+      ],
+      Array [
+        -77.923104,
+        42.096481,
+      ],
+    ],
+  ],
+  "updatedControlPoints": Array [
```
evansiroky: This snapshot makes no sense due to bad input data. The pattern coordinates are in NY while the control point coordinates are in CA.

landonreed:
Yes, sorry about that. I was trying to adapt preexisting fixtures and that was a bad idea. I'll regenerate them from new input data.

lib/editor/util/__tests__/__snapshots__/map.js.snap
```diff
@@ -834,96 +665,40 @@ Object {
...
-      ],
-      Array [
-        -77.922974,
-        42.095657,
-      ],
-      Array [
-        -77.923104,
-        42.096481,
+        -77.92187333106995,
+        42.08646327816015,
```
evansiroky: This coordinate in the snapshot for `should update the first stop` seems out of place. This indicates that the shape goes backwards.

Here are the pattern coordinates and stops before the test:

Screen Shot 2019-04-03 at 2 53 02 PM
 
And after recalcuate shape is ran:

Screen Shot 2019-04-03 at 2 53 13 PM

The segment between the first and 2nd stop looks fine, but the shape doesn't continue to the last stop. Instead it goes backwards. You can't see if from that last picture, so here is a more zoomed in look:

Screen Shot 2019-04-03 at 3 04 43 PM

lib/editor/util/__tests__/__snapshots__/map.js.snap
```diff
@@ -1030,76 +797,32 @@ Object {
   "coordinates": Array [
     Array [
       Array [
-        -77.922203,
-        42.086407,
+        -77.9230535030365,
+        42.096176462786296,
```
evansiroky: The same odd thing in `should update the first stop` is now happening in reverse in the `should update the last stop`.

Before the test:

Screen Shot 2019-04-03 at 3 11 40 PM

After:

Screen Shot 2019-04-03 at 3 11 23 PM

Zoomed in the first point is now going backwards to some other point and going forward.

Screen Shot 2019-04-03 at 3 13 19 PM

lib/editor/util/__tests__/fixtures/test-control-points.json
```diff
@@ -40,16 +38,12 @@
...
-  "defaultDwellTime": 0,
-  "timepoint": null,
-  "shapeDistTraveled": 691.729078104771,
-  "hidden": true
+  "stopId": "a9c9cd55-ca3f-4ce7-9a93-96f2f4eaf089"
 }, {
   "distance": 905.1384059344052,
   "id": "0cj3",
-  "permanent": true,
+  "pointType": 2,
```
evansiroky: This should probably be `1` since there is no stopId for this control point.

lib/editor/util/__tests__/__snapshots__/map.js.snap
```diff
@@ -583,46 +448,22 @@ Object {
...
-      ],
-      Array [
-        -77.922196,
-        42.086498,
-      ],
-      Array [
-        -77.922173,
-        42.086498,
+        -77.9230535030365,
+        42.096176462786296,
```
evansiroky: This snapshot for `should update a point in the middle of the control points` seems really off again.

Before:

Screen Shot 2019-04-03 at 4 41 44 PM

After:

Screen Shot 2019-04-03 at 4 41 35 PM

lib/editor/util/__tests__/__snapshots__/map.js.snap
```diff
@@ -160,70 +125,30 @@ Object {
   "coordinates": Array [
     Array [
       Array [
-        -77.922135,
-        42.089099,
```
evansiroky: The result of the `should delete the first stop` test implies that the last stop is deleted and the line is backwards!

Before:

Screen Shot 2019-04-03 at 5 06 01 PM

After:

Screen Shot 2019-04-03 at 5 05 52 PM

landonreed:
Agh, you're right. Sorry about that. I'm going to add some log statements that actually print out geojson.io links to make this stuff a little easier to debug.

PR #402, opened 2019-01-25
Home page refactor

landonreed:
For a long time the landing page for the application has been just this empty leaflet map and a dropdown selector that doesn't work. This has been really bugging me, so this PR primarily seeks to refresh the landing page. Other things it does:

- Makes things on the landing page (title, description, logo) configurable
- Removes some older components for public viewing of pages that we're not using
- Restyles the user buttons (my account, admin, log out)
- Refreshes things with a new application-wide font (Google's [Sarabun](https://fonts.google.com/specimen/Sarabun))

### Logged out
image

### Logged in
image


lib/public/components/PublicLandingPage.js
```diff
@@ -0,0 +1,86 @@
...
+                {!this.props.user.profile
+                  ? <span>Existing users <Link to='/login'>sign in here</Link>.</span>
+                  : null
+                }
+              </p>
+            </Col>
+          </Row>
+        </Grid>
+        <footer
+          className='footer'
```
evansiroky: This className does not appear to be necessary... unless you create an entry in styles.css and put the style on the following lines there.  Please either remove the className or move the style info into the stylesheet file.

PR #409, opened 2019-02-06
Make sure the status message is cleared when updating users

### LINKED ISSUE #408: When updating a user other than the currently logged in user, the status message will not be cleared properly
## Steps to reproduce:

In the admin screen, try updating a user other than the account that is currently logged in.

## Problem behavior:

Status message of `Updating user...` does not get cleared after a successful request.

## Expected behavior:

Status message of `Updating user...` should get cleared upon the request finishing.

evansiroky:
fixes #408

### DIFF
--- a/lib/manager/actions/user.js
+++ b/lib/manager/actions/user.js
@@ -15,6 +15,7 @@ import type {RecentActivity, UserProfile} from '../../types'
 import type {dispatchFn, getStateFn, ManagerUserState} from '../../types/reducers'
 
 const checkingExistingLogin = createVoidPayloadAction('CHECKING_EXISTING_LOGIN')
+const clearStatusMessage = createVoidPayloadAction('CLEAR_STATUS_MESSAGE')
 const receiveRecentActivity = createAction(
   'RECEIVE_USER_RECENT_ACTIVITY',
   (payload: Array<RecentActivity>) => payload
@@ -256,6 +257,12 @@ export function updateUserData (user: any, userData: any) {
           // If user being updated matches the logged in user, update their
           // profile in the application state.
           dispatch(userProfileUpdated(updatedProfile))
+        } else {
+          // a different user than the current user was updated. An action needs
+          // to be dispatched because the status message from the
+          // updatingUserData is still set to something, so we need to indicate
+          // that the request completed.
+          dispatch(clearStatusMessage())
         }
       })
   }
--- a/lib/manager/actions/user.js
+++ b/lib/manager/actions/user.js
@@ -223,7 +223,10 @@ export function updateTargetForSubscription (profile: UserProfile, target: strin
 }
 
 /**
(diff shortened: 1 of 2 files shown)


lib/manager/actions/user.js
```diff
@@ -256,6 +257,12 @@ export function updateUserData (user: any, userData: any) {
           // If user being updated matches the logged in user, update their
           // profile in the application state.
           dispatch(userProfileUpdated(updatedProfile))
+        } else {
+          // a different user than the current user was updated. An action needs
+          // to be dispatched because the status message from the
+          // updatingUserData is still set to something, so we need to indicate
+          // that the request completed.
+          dispatch(clearStatusMessage())
```
landonreed: I wonder if rather than clearing the status message, we might want to replace the user profile that has been updated with the one contained in the store or alternatively, just fetch the user list again. Either way I think updating the user with its up-to-date info so that the user can verify the update matches their desired change is important.

evansiroky:
I refactored this so that an error message is shown when the update fails and that the user list is refreshed upon success.

evansiroky:
The new code changes the behavior after a successful update so that the user list is refreshed each time. Therefore, the check to update is no longer needed, because the most recent data gets refetched.

That error message was copy-pasta-no-updata. Pushed a fix for that.

evansiroky:
Just added back logic to update current user permissions after a self-update is performed.

lib/manager/actions/user.js
```diff
@@ -242,21 +243,31 @@ export function updateUserData (user: any, userData: any) {
...
-        if (updatedProfile.user_id === currentUserId) {
+        if (responseJson.user_id === profile.user_id) {
           // If user being updated matches the logged in user, update their
-          // profile in the application state.
-          dispatch(userProfileUpdated(updatedProfile))
+          // profile in the application state as the change in permissions
+          // may affect what they're allowed to do in the application.
+          dispatch(userProfileUpdated(responseJson))
         }
+        dispatch(fetchUsers())
```
landonreed: I don't think `fetchUsers` should always be called. This function is called in other places besides the user admin module (e.g., `updateTargetForSubscription`). So when a user updates their own subscriptions, we really don't want to be fetching all users. This was the original purpose of the code that updates the logged in user that was removed in a previous commit (and was added back here). Given how much confusion has surrounded this relatively minor code change, I think you should update this method's doc to read:
```
Update application/client ID specific datatools object with provided user data. This is used in the user admin module to update some user's permissions as well as throughout the application to update attributes for the logged in user (e.g., when a user subscribes/watches a feed source or project).

evansiroky: Oh wow, yeah, I didn't realize how widely this action was used and how often times it's for updating subscriptions and other non-permissions stuff.

evansiroky:
I added more comments and refactored as suggest.

PR #460, opened 2019-07-31
Fix #452 by use of Local Storage

### LINKED ISSUE #452: Auth0 returns a stale userInfo
## Observed behavior 
Watch/Unwatch button doesn't mantain the correct state; this is because data retrieved from auth0 /userinfo call is not updated with last modifications, despite app_metadata on auth0 user profile gets correctly updated

## Expected behavior
UI gets updated with app_metadata modifications

- click watch button on FeedViewer manager page (the app_metadata gets updated on Auth0)
- navigate to the feed's project page (/userinfo is called BUT retrieve stale data)
- navigate back to the feed page: the button still show "watch" even if it should show "unwatch"

We found this enlightening thread on Auth0 forum: 
[https://community.auth0.com/t/auth0-returns-a-stale-userinfo/13030/6](https://community.auth0.com/t/auth0-returns-a-stale-userinfo/13030/6)
TLDR: /userinfo endpoint retrieve historical data from the latest authentication point in time

## Version of datatools-ui and datatools-server if applicable 
- dev brench on both server and ui
- using a new Auth0 tenant

aguiraf:
### Checklist

### Description

Store, update and retrieve user profile data using Local Storage. Seeks to resolve #452. 
- Auth0 /userinfo endpoint is called only during login phase, profile data is stored on Local Storage.
- any update on user profile is sent to Auth0 AND to Local Storage
- subsequent calls to userinfo are bypassed (data would be stale) and profile is returned from Local Storage
- During the logout phase the local storage entry is wiped

This is intended for patching the Auth0 unespected behaviour until the new work that will use MongoDB to store user information.

### DIFF
--- a/lib/manager/actions/user.js
+++ b/lib/manager/actions/user.js
@@ -273,6 +273,9 @@ export function updateUserData (user: any, userData: any) {
         } else {
           dispatch(fetchUsers())
         }
+        // We need to update directly localStorage as Auth0 returns stale data
+        // see https://github.com/ibi-group/datatools-ui/issues/452
+        window.localStorage.setItem('profile', JSON.stringify(profile))
       })
   }
 }
--- a/lib/manager/actions/user.js
+++ b/lib/manager/actions/user.js
@@ -275,7 +275,7 @@ export function updateUserData (user: any, userData: any) {
         }
         // We need to update directly localStorage as Auth0 returns stale data
         // see https://github.com/ibi-group/datatools-ui/issues/452
-        window.localStorage.setItem('profile', JSON.stringify(profile))
+        auth0.setProfile(profile)
       })
   }
 }
--- a/lib/common/user/Auth0Manager.js
+++ b/lib/common/user/Auth0Manager.js
@@ -59,10 +59,10 @@ class Auth0Manager {
 
   /**
    * Use the access token (token for API calls) to get the user's profile.
-   * We can't fetch the profile data all the time because we'll get 429'ed by
-   * Auth0.  Therfore, we also store the profile data in localStorage, but we
-   * assume that the profile data is only valid for 60 seconds since it is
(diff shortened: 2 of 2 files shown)


lib/manager/actions/user.js
```diff
@@ -273,6 +273,9 @@ export function updateUserData (user: any, userData: any) {
         } else {
           dispatch(fetchUsers())
         }
+        // We need to update directly localStorage as Auth0 returns stale data
+        // see https://github.com/ibi-group/datatools-ui/issues/452
+        window.localStorage.setItem('profile', JSON.stringify(profile))
```
evansiroky: It'd probably be best to use `Auth0Manager.setProfile` to avoid duplicating code.

evansiroky:
After seeing this proposed code and the problem description in #452, I think this raises a security issue. It seems that it is possible for an admin to revoke permissions for a particular user, but since all of the app_metadata is stale while appearing valid to the backend it is possible for the user with revoked permissions to perform unauthorized actions as long as they're logged in on the website - both currently and with these changes. As @landonreed mentioned, we do hope to refactor datatools to store all user data outside of Auth0 in a backend DB (ie Mongo) and always be fetching the latest data from that backend DB to check the user's authorizations before completing certain actions. This PR does "fix" the issue with stale data for individual user accounts, but a larger refactor is needed to ensure that user data is updated after external events cause changes in user data (ie when an admin updates another users' data).

landonreed:
As @evansiroky mentioned, we're looking to refactor how user data is stored soon. This change will affect how users subscribe to watch feeds and projects, so for now we're going to close this. Thanks for reporting the issue and providing the fix though. We'll keep you updated on the user data storage update progress.

aguiraf:
thank you, looking forward mongodb solution!

PR #463, opened 2019-08-07
Deploy to ec2/Manage OTP servers at the application level

landonreed:
### Checklist

### Description

Note: this code was taken from catalogueglobal/datatools-ui#256

This is the UI PR to accompany ibi-group/datatools-server#225. It moves the management of OTP servers from the project settings to the site admin settings and allows servers to be optionally available to all projects in an application (or project specific). This also resolves catalogueglobal/datatools-ui#234, another deployment-related ticket.

This also fixes the merge project feeds job to conform with the standard approach of attaching just the ID of the related entity (project in this case) to the job JSON object, rather than the whole project object.

To test out the actual deployment @evansiroky, you can create an OTP server in the app at http://localhost:9966/admin/servers based on the following JSON I copied from my Mongo instance:

```
{
    "admin" : false,
    "instanceType" : "r4.xlarge",
    "internalUrl" : [],
    "name" : "IBI Test",
    "publicUrl" : "ibi-test-otp-828479082.us-east-1.elb.amazonaws.com",
    "targetGroupArn" : "arn:aws:elasticloadbalancing:us-east-1:739199387312:targetgroup/ibi-test-otp/c28d4b2ff45fe3e2",
    "projectId" : null,
    "s3Bucket" : "datatools-production"
}
```


i18n/espanol.yml
```diff
@@ -1,738 +0,0 @@
-_id: es
```
evansiroky: Adiós español.

lib/admin/components/UserAdmin.js
```diff
@@ -1,59 +1,76 @@
...
+  fetchServers: typeof adminActions.fetchServers,
   fetchUsers: typeof adminActions.fetchUsers,
   isFetching: boolean,
   organizations: OrganizationsState,
+  otpServers: Array<OtpServer>,
   projects: Array<Project>,
   setUserPage: typeof adminActions.setUserPage,
   setUserQueryString: typeof adminActions.setUserQueryString,
   updateOrganization: typeof organizationActions.updateOrganization,
+  updateServer: typeof adminActions.updateServer,
```
evansiroky: Way too many props that aren't needed directly in this component. This needs a refactor so that some of the child components in this component become connected components so that we can avoid passing through so many props.

lib/admin/components/UserAdmin.js
```diff
@@ -66,30 +83,104 @@ export default class UserAdmin extends Component<Props> {
...
       user,
       users
     } = this.props
+    switch (activeComponent) {
+      case 'users':
+        if (!users.data || !projects || !organizations.data) {
+          return <Loading />
+        } else {
+          return (
+            <UserList
```
evansiroky: So many props... should be a connected component.

lib/admin/components/UserAdmin.js
```diff
@@ -66,30 +83,104 @@ export default class UserAdmin extends Component<Props> {
...
+              token={user.token || 'token-invalid'}
+              updateUserData={updateUserData}
+              userCount={users.userCount}
+              users={users.data}
+            />
+          )
+        }
+      case 'organizations':
+        if (!isApplicationAdmin || isModuleEnabled('enterprise')) return null
+        else return <OrganizationList {...this.props} />
```
evansiroky: Probably sending even more props than are actually needed. The OrganizationList should be a connected component.

lib/admin/components/UserAdmin.js
```diff
@@ -66,30 +83,104 @@ export default class UserAdmin extends Component<Props> {
...
+            bsSize='large'
+            href='https://manage.auth0.com/#/logs'>
+            <Icon type='star' /> View application logs on Auth0.com
+          </Button>
+        </p>
+      case 'servers':
+        if (!isApplicationAdmin || !isModuleEnabled('deployment')) {
+          return <p className='text-center lead'>Restricted access</p>
+        }
+        return <ServerSettings
```
evansiroky: Not too many props on this one, but should probably also be a connected component.

lib/manager/components/CollapsiblePanel.js
```diff
@@ -17,12 +18,14 @@ import {getComponentMessages} from '../../common/util/config'
 import type {FormProps} from '../../types'
 
 type Props = {
-  data: {[string]: string | number},
+  children?: Element<*>,
+  data: any,
```
evansiroky: Would be nice to have a more detailed type here.

lib/manager/components/DeploymentViewer.js
```diff
@@ -102,6 +111,33 @@ export default class DeploymentViewer extends Component<Props, State> {
     return [[north, east], [south, west]]
   }
 
+  _loadOptions = () => {
+    fetch('https://r5-builds.s3.amazonaws.com')
```
evansiroky: This should probably be a config value instead of being hardcoded.

lib/manager/components/DeploymentViewer.js
```diff
@@ -102,6 +111,33 @@ export default class DeploymentViewer extends Component<Props, State> {
...
+        const parser = new window.DOMParser()
+        const r5doc = parser.parseFromString(text, 'application/xml')
+
+        const all = Array.from(r5doc.querySelectorAll('Contents'))
+          .map(item => item.querySelector('Key').childNodes[0].nodeValue) // get just key
+          .filter(item => item !== 'index.html') // don't include the main page
+          .map(item => item.replace(/.jar$/, '')) // and remove .jar
+        this.setState({r5: all})
+      })
+    fetch('https://opentripplanner-builds.s3.amazonaws.com')
```
evansiroky: This should probably be a config value instead of being hardcoded.

lib/manager/components/DeploymentViewer.js
```diff
@@ -102,6 +111,33 @@ export default class DeploymentViewer extends Component<Props, State> {
...
+      .then(res => res.text())
+      .then(text => {
+        const parser = new window.DOMParser()
+        const r5doc = parser.parseFromString(text, 'application/xml')
+
+        const all = Array.from(r5doc.querySelectorAll('Contents'))
+          .map(item => item.querySelector('Key').childNodes[0].nodeValue) // get just key
+          .filter(item => item !== 'index.html') // don't include the main page
+          .map(item => item.replace(/.jar$/, '')) // and remove .jar
+        this.setState({otp: all})
```
evansiroky: Lots of very similar code in here and for r5. Can it be combined?

lib/manager/components/DeploymentViewer.js
```diff
@@ -168,6 +216,9 @@ export default class DeploymentViewer extends Component<Props, State> {
       {east: 10, north: 10, west: -10, south: -10}
     const boundsTooLarge = east - west > BOUNDS_LIMIT || north - south > BOUNDS_LIMIT
     const bounds = this._getBounds()
+    const serverDeployedTo = deployment.deployedTo
```
evansiroky: Would prefer having this be a boolean variable instead of possibly null.

lib/manager/components/DeploymentViewer.js
landonreed: Do you think the bucket name or the entire URL should be configurable? I'd lean towards bucket name.

lib/manager/components/DeploymentViewer.js
landonreed: Can you clarify? Are you referring to `serverDeployedTo`? This variable (type `OtpServer`) is used below to get the name of the server.

evansiroky: Oh, whoops, nevermind.

evansiroky:
These buttons are small and I didn't see them at first since they were in the upper right. I kinda would prefer for them to be larger, have the action spelled out and be on the bottom of the card for the server form.

Screen Shot 2019-08-14 at 12 25 34 PM

lib/types/index.js
```diff
@@ -518,10 +521,14 @@ export type Profile = {
...
-  internalUrl: string,
+  id: string,
+  instanceType: ?string,
+  internalUrl: ?string,
   name: string,
+  projectId: ?string,
   publicUrl: string,
-  s3Bucket: ?string
+  s3Bucket: ?string,
+  targetGroupArn: ?string
```
evansiroky: This is missing `instanceCount` field that is present in the datatools-server PR.

lib/admin/components/AdminPage.js
```diff
@@ -0,0 +1,181 @@
...
+import {connect} from 'react-redux'
+import Icon from '@conveyal/woonerf/components/icon'
+import * as React from 'react'
+import {
+  Grid,
+  Row,
+  Col,
+  Panel,
+  ListGroup,
+  ListGroupItem as LGI,
```
evansiroky: We shouldn't abbreviate component names.

lib/admin/components/ServerSettings.js
```diff
@@ -0,0 +1,269 @@
...
+    this._updateStateFromProps(nextProps)
+  }
+
+  _updateStateFromProps = props => {
+    if (props.otpServers && props.otpServers.data) {
+      this.setState({ otpServers: props.otpServers.data })
+    }
+  }
+
+  _getOnChange = (evt, index) => {
```
evansiroky: This is beginning to look like we're inventing a React form framework. This isn't a good place for that because it's within just this component. We should not have every single component have their own form frameworks and instead just pick one and use it everywhere to be consistent. I liked using [React-Formal](http://jquense.github.io/react-formal/) in the commute project.

lib/admin/reducers/servers.js
```diff
@@ -0,0 +1,32 @@
...
+
+import type { Action } from '../../types/actions'
+import type { AdminServersState } from '../../types/reducers'
+
+export const defaultState = {
+  isFetching: false,
+  data: null
+}
+
+const users = (state: AdminServersState = defaultState, action: Action): AdminServersState => {
```
evansiroky: copy-pasta: `users = ...`

evansiroky: Sad, we still don't have flow typing for actions in the reducers.

lib/admin/components/AdminPage.js
```diff
@@ -0,0 +1,181 @@
...
+              </div>
+            }
+          </Row>
+        </Grid>
+      </ManagerPage>
+    )
+  }
+}
+
+const mapStateToProps = (state: AppState, ownProps: Props) => {
```
evansiroky: `Props` is not the right props here. That is for the component that is connected. `ownProps` should be a `RouterProps` type.

lib/admin/components/ServerSettings.js
```diff
@@ -0,0 +1,269 @@
...
+
+import { getComponentMessages } from '../../common/util/config'
+import * as adminActions from '../actions/admin'
+import CollapsiblePanel from '../../manager/components/CollapsiblePanel'
+import { FIELDS, SERVER_FIELDS, UPDATER_FIELDS } from '../../manager/util/deployment'
+
+import type { OtpServer, Project } from '../../types'
+import type { AppState, AdminServersState } from '../../types/reducers'
+
+type ContainerProps = {editDisabled: boolean}
```
evansiroky: I think this should be defined or inlined down where it's used.

lib/manager/actions/status.js
```diff
@@ -194,6 +199,15 @@ export function handleFinishedJob (job: ServerJob) {
         break
       case 'DEPLOY_TO_OTP':
         return dispatch(fetchDeployment(job.deploymentId))
+      // No completion step for build transport network. User should just
+      // re-request isochrones.
+      case 'SYNC_PROJECT_FEEDS':
```
evansiroky: How does this action ever get fired?

landonreed: Oi, I think the branch I pulled in from catalogueglobal must have also had some changes related to https://github.com/catalogueglobal/datatools-server/pull/32/files#diff-42c8506cd760bd0551bfefeb52ce5c33R23.

lib/admin/components/ServerSettings.js
landonreed: Good point. Much of this form handling code was taken out of `DeploymentSettings`. Do you think we should create a separate issue/PR for discussing/implementing some other form framework?

lib/admin/components/ServerSettings.js
landonreed: It's actually used in the line directly below (in addition to the very bottom of the file).

landonreed:
Thanks for the review, @evansiroky. See https://github.com/ibi-group/datatools-ui/pull/463#discussion_r320348011. Also, could you either post your error message here or maybe we should do a screenshare?

lib/admin/components/ServerSettings.js
evansiroky: Yeah, would be good to create an issue for that.

evansiroky:
I'd definitely like to do a screenshare.

lib/admin/components/ServerSettings.js
evansiroky: A, makes sense. I recall importing the container props from another container file, but now I see that the redux connection is in this file also, so this makes sense.

lib/admin/components/ServerSettings.js
```diff
@@ -0,0 +1,403 @@
...
+  }
+
+  _onAddServer = () => {
+    const newServer = {
+      name: '',
+      publicUrl: '',
+      internalUrl: [],
+      admin: false
+    }
+    const stateUpdate = { otpServers: { $push: [newServer] } }
```
evansiroky: This variable can probably just be inlined in the update call below.

lib/admin/components/ServerSettings.js
```diff
@@ -0,0 +1,403 @@
...
+              these servers, but the below instructions detail the generally
+              recommended approaches.
+            </p>
+            <h4>Production deployments</h4>
+            <p>
+              The recommended approach for production deployments is to create
+              an elastic load balancer (ELB) in AWS along with a target group.
+              You should then reference the target group ARN in the server
+              (an S3 bucket must be identified as well). The steps for this
+              approach are as follows:
```
evansiroky: There could definitely be more documentation for this process and how to fill in the fields appropriately. Better yet, maybe there could be some kind of server-side logic that would automatically setup a load balancer.

lib/common/components/EC2InstanceCard.js
```diff
@@ -0,0 +1,106 @@
...
+    // If instance hasn't launched, don't calculate time.
+    if (!instance.launchTime) return 'N/A'
+    // Set end time as termination time or now.
+    const endTimestamp = this._getTerminationTime()
+    let endTime = endTimestamp ? moment(endTimestamp) : moment()
+    const launchTime = moment(instance.launchTime)
+    const duration = moment.duration(endTime.diff(launchTime))
+    return duration.asDays() < 1
+      ? `${duration.asHours().toFixed(2)} hours`
+      : `${duration.asDays().toFixed(2)} days`
```
evansiroky: There's already the `humanize-duration` package that is dependency for datatools that could do this.

lib/common/components/EC2InstanceCard.js
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  _getStatusIcon = (instance: EC2InstanceSummary) => {
+    // 'pending' | 'running' | 'shutting-down' | 'stopping' | 'stopped' | 'terminated'
+    const color = instance.state.name === 'running'
+      ? 'green'
+      : instance.state.name === 'pending' || instance.state.name === 'shutting-down'
+        ? 'yellow'
+        : 'red'
+    return <Icon title={instance.state.name} type='circle' style={{ color }} />
```
evansiroky: Might be good to also include the text of the status

lib/common/components/FormInput.js
```diff
@@ -0,0 +1,70 @@
...
+  render () {
+    const {field, messageRoot, shouldRender, value} = this.props
+    const prefix = messageRoot ? `${messageRoot}.` : ''
+    const {effects, split, ...fieldProps} = field
+    // If should render is false, do not render form element (for conditional
+    // visibility of form elements).
+    if (!shouldRender()) return null
+    const isCheckbox = field.type === 'checkbox'
+    const isSelect = field.type.indexOf('select') !== -1
+    const checkboxProps = isCheckbox ? {checked: value} : {}
```
evansiroky: Definitely looking like a new form framework. Gonna pitch [React Formal](http://jquense.github.io/react-formal/) again.

lib/manager/components/DeploymentViewer.js
```diff
@@ -348,8 +402,28 @@ export default class DeploymentViewer extends Component<Props, State> {
...
+                    checked={deployment.r5}
+                    onChange={this._onChangeR5}>Use R5</Checkbox>
+                  <Checkbox
+                    checked={deployment.buildGraphOnly}
+                    onChange={this._onChangeBuildGraphOnly}>Build graph only</Checkbox>
+                  {deployment.r5 ? 'R5' : 'OTP'} version
+                  <Select
+                    value={deployment.r5 ? deployment.r5Version : deployment.otpVersion}
+                    onChange={this._onUpdateVersion}
+                    options={options ? options.map(v => ({value: v, label: v})) : []} />
```
evansiroky: Would be nice to set the build bucket so we can use OTP builds from the IBI fork.

lib/manager/components/DeploymentsPanel.js
```diff
@@ -137,17 +143,32 @@ class DeploymentsList extends Component<ListProps, State> {
...
+          if (!a.name) return -1
+          if (!b.name) return 1
+          // Ensure pinned deployment is first element in list.
+          if (b.id === pinnedDeploymentId) return 1
+          if (a.id === pinnedDeploymentId) return -1
+          // Otherwise, sort by most recent last deployed date and then most recent
+          // created date (ensuring deployments never deployed show up at end of
+          // list).
+          // TODO: Refactor to allow for sorting on multiple fields? This may be
+          //  overkill for deployments.
```
evansiroky: I agree this may be a bit much to have multi-level sorting. Perhpas it would be possible to implement a sorting button like was done for the project feeds table.

lib/manager/components/CurrentDeploymentPanel.js
```diff
@@ -0,0 +1,105 @@
...
+    const ec2Info = server && server.ec2Info
+    return (
+      <Panel header={<h3><Icon type='server' /> Current deployment</h3>}>
+        {deployment.deployedTo
+          ? <div>
+            <ul className='list-unstyled'>
+              <li>
+                Deployed to: {this._getServerLabel()}
+              </li>
+              <li>Using ELB?: {ec2Info ? 'Yes' : 'No (built graph over wire)'}</li>
```
evansiroky: Since everything below here is so dependent on using an ELB, I think it may be useful to have 3 distinct types of deployments:

- deploy just the bundle to s3
- build graph over wire deployment
- ELB/EC2 deployment

In doing this, we can also modify the forms such that only the needed fields for each deployment type are shown.

lib/admin/components/ServerSettings.js
landonreed: Could you suggest some changes to the documentation, since you went through the process of setting it up somewhat blind?

Regarding the auto-setup by the server, I want to be careful not to over-automate this. We likely need this to be flexible enough to work with across AWS accounts/networks and I'm not sure if there is a great way to account for that with an automatic procedure. For example, if we have Transit Agency X that wants IBI prod Data Tools to deploy to an ELB in their own AWS account, we can set up the IAM permissions accordingly, but there may be too many factors to handle an auto-setup.

lib/common/components/EC2InstanceCard.js
landonreed: Ah, good call.

lib/common/components/FormInput.js
landonreed: For this PR, I don't think it's a good idea to replace the form framework entirely. I'm just working within the confines of what is already set up (I did do some refactoring/simplification of the form components). But I do agree that it might be a good call to replace this soon.

lib/manager/components/DeploymentViewer.js
landonreed: As in, set the build bucket in the config file or by some other means? Do we have a way to write IBI builds to a bucket?

lib/manager/components/DeploymentsPanel.js
landonreed: Agreed, I'll create an issue from this comment.

lib/manager/components/CurrentDeploymentPanel.js
```diff
@@ -0,0 +1,105 @@
...
+              <li>Last deployed: {formatTimestamp(deployment.lastDeployed)}</li>
+            </ul>
+            <div style={{ margin: '20px 0' }}>
+              <DeploymentPreviewButton block deployment={deployment} />
+            </div>
+          </div>
+          : deployJob
+            ? <span>Deployment to {this._getServerLabel()} in progress...</span>
+            : 'No current deployment found'
```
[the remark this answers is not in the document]
landonreed: - Agree on status.
- I like the idea of being able to cancel a deployment job (or any MonitorableJob for that matter), but I think that would require us to change some things around with the `DataManager.heavyExecutor`. Basically, I think we would need that to be a `ScheduledExecutorService` and handle job kickoff/storage a little differently.
- I'm going to add some changes that allow you to download the build artifacts. to address the other items.

lib/common/components/EC2InstanceCard.js
```diff
@@ -0,0 +1,106 @@
...
+              <a
+                style={{ marginLeft: '60px' }}
+                target='_blank'
+                href={this._getInstanceUrl()}>view in console</a>
+            </div>
+            {terminationTime
+              ? <div><Icon type='times' /> {formatTimestamp(terminationTime)}</div>
+              : null
+            }
+          </small>
```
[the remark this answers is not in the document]
landonreed: @evansiroky, I'm going to hold off on manually terminating a single instance. I'm not sure there's a strong need for this feature. We can just do this with the AWS console or normal users can effectively do this by re-deploying. I've made some improvements in this latest refactor that should make server startup failure less likely.

landonreed:
I've made a number of changes/improvements. Thanks for the great feedback!

One thing you'll need to change on your end is updating the OtpServer field name in MongoDB from `ec2Info.iamRoleArn -> ec2Info.iamInstanceProfileArn`. Other than that, I think all you need to do is click Deploy and test out the changes!

lib/manager/components/DeploymentViewer.js
evansiroky: Yeah, we should probably be writing IBI OTP builds to a separate bucket. Not sure how we'd add that bucket to datatools. Might be a good idea to have it be a config item I suppose.

evansiroky:
While these changes are a step in the right direction and do provide _raw_ information about what the ec2 instances are running, I still think we should rethink the design of deployments as noted [here](https://github.com/ibi-group/datatools-ui/pull/463#issuecomment-531113941) and [here](https://github.com/ibi-group/datatools-ui/pull/463#discussion_r324380804). Right now, when I go to view a deployment, the amount of screen real estate devoted to information about the active deployment is very small compared to the map and other items which if changed have no effect on the active deployment. Furthermore, the raw information about the bundle is helpful, but the details about what feed versions were used are hard to retrace.

Also, in my testing of deploying 10 instances, basically every time an I tried, there was some kind of failure. I still think there are more opportunities to implement more detailed fault tolerance with automatic retrying and for better progress reporting so that it is easier for a user to understand what's going on and debug what went wrong. Ideally, we'd have more detailed logs that we can be automatically parsed to examine where failures occurred so that we know where to focus our efforts and detect any regressions. 

Also, there's no way to add an extra instance to a deployment, which I can imagine might be desired if not all of the instances start up correctly. If we're trying to get something shipped fast, then we can probably merge this once the Travis builds pass, but we should be prepared to tell clients that things aren't perfect and train them a bit on how the UI works.

lib/manager/components/CurrentDeploymentPanel.js
landonreed: These are not mutually exclusive types though. For example, you can build the graph over the wire and deploy to s3. Also, the S3 bucket field is necessary for ELB. At the moment this UI is essentially just limited to application admins (i.e., us), so while I think it is important to improve the UI of this form, I think that deserves a separate conversation at this point.

landonreed:
Can you describe more fully all of the errors you encountered when starting up the 10 instances? Also, what commits/feeds were you using? Yesterday, I pushed some more commits due to some issues I faced with the FDOT instance that may fix a few things.

I think our goal here for this PR should be to eliminate the failing instances you're encountering while starting up. The goal for this PR is to introduce this new functionality that dramatically reduces the risk for downtime due to deploying with build graph over wire. I understand your interest in seeing additional features like starting up/terminating single instances, but my perspective on that remains that for now that can entirely be handled by re-deploying to the ELB. Sure, that may change after we've used this feature for a bit, but I don't want to over-speculate on what needs to be in here and spend time building out unneeded features.

lib/manager/components/CurrentDeploymentPanel.js
evansiroky: I think it would make things simpler if those 3 items were mutually exclusive. Anyways, this will be something to explore as we get feedback from clients.

lib/admin/components/ServerSettings.js
evansiroky: Making these docs into a step-by-step guide with screenshots in the documentation is what I would've liked to see. And regarding the automatic ELB setup, I was thinking it could be a good default, but that option would require more maintenance and support.

evansiroky:
Each time, there was a different error, unfortunately. But I just tried again this morning and everything went smoothly. Most of the time, the errors seemed to be interruptions in uploading or downloading stuff. But based on my experience, I think we should be prepared for our clients to also see various random errors. 

I still also think the UI could use a redesign to have more rigidly defined deployment types and to better distinguish between active deployments and deployment templates, but if we really want to get this shipped and improve those items later then we can merge.

evansiroky:
Travis builds need to pass, but should be good to go after that.

landonreed:
@evansiroky, any ideas on why the snapshots are failing here but passing on our local machines?

PR #467, opened 2019-08-21
Migrate user script improvements

evansiroky:
### Checklist

### Description

The original migrate users script was not preserving the original `user_id` data for users. This script corrects that. Also, the script will now check if a user already exists in a tenant and will (by default) try to delete that user and retry creation.

### DIFF
--- a/scripts/migrate-auth0-users.js
+++ b/scripts/migrate-auth0-users.js
@@ -7,10 +7,21 @@
  *
  * Two config files are required for this to work. The `from-client.json` will
  * contain information for machine-to-machine connections and the existing
- * client application id that users should have in their metadata. The
+ * client application ID that users should have in their metadata. The
  * `to-client.json` file should have information for machine-to-machine
- * connections and the client application id that will replace to old
- * applicaiton client id in their metadata. Each of these files should have the
+ * connections and the client application ID that will replace to old
+ * applicaiton client ID in their metadata.
+ *
+ * If deleting (and then recreating) existing users in the to tenant is desired,
+ * the connection ID is also required. A way to find the connection ID involves
+ * going to the Auth0 API explorer, setting the API token with a token generated
+ * from a machine-to-machine API for the to tenant and then executing the "Get
+ * all connections" endpoint.
+ * See here: https://auth0.com/docs/api/management/v2#!/Connections/get_connections.
+ * That endpoint will give a listing of all connections. Find the connection
+ * with the name "Username-Password-Authentication" and use that ID.
+ *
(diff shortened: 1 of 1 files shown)


scripts/migrate-auth0-users.js
```diff
@@ -20,17 +20,25 @@
  *   "clientId": "SECRET"
  * }
  *
+ * If deleting existing users in the to tenant id desired, the connection id is
+ * also required. The connection id typically is only discoverd by opening up
+ * an API explorer and listing all connections.
+ *
```
landonreed: I feel like this belongs above the example format for `client.json` in a parenthetical. Also, I think this needs a bit more information about how to find the `connectionId` and it should be listed in the example base format (probably with a single line comment saying it's only required for user deletion).

scripts/migrate-auth0-users.js
```diff
@@ -66,6 +74,75 @@ async function getManagementToken (connectionSettings) {
...
+      headers: makeHeaderWithToken(token),
+      method: 'POST'
+    }
+  )
+
+  // check if request succeeded
+  const createFailed = response.status >= 400
+  const responseJson = await response.json()
+  if (createFailed) {
+    // althought the creation failed, don't raise an error as most
```
landonreed: althought typo

scripts/migrate-auth0-users.js
```diff
@@ -78,107 +155,97 @@ async function main () {
...
+          'last_login',
+          'last_password_reset',
+          'created_at',
+          'updated_at',
+          'identities'
+        ]
+      )
+
+      // modify the user_id so an extra prefix isn't created
+      user.user_id = user.user_id.substr(6)
```
landonreed: This might need a little extra explanation. Is this where the user_id is preserved? Maybe just expand a little bit on where this extra prefix comes from and why it's discarded.

scripts/migrate-auth0-users.js
```diff
@@ -78,107 +155,97 @@ async function main () {
...
+
+      // attempt to create the user
+      const createUserResult = await createUser(to.domain, toToken, user)
+      if (!createUserResult.success) {
+        // creation of user not successful, check if script should try again by
+        // deleting an existing user
+        if (
+          createUserResult.responseJson.message === 'The user already exists.' &&
```
landonreed: Is there a more stable way to check for this? I'm guessing not, since the HTTP status code is probably also used in other cases, but just wanted to check.

evansiroky: Yeah, not really. I think this message was the most specific part of the response.

evansiroky:
Ready for re-review.

PR #470, opened 2019-08-23
refactor: create better wording for bounds placeholders

### LINKED ISSUE #471: The bounds field in the DeploymentSettings component doesn't save bounds properly

Whenever custom bounds are attempted to be entered in the DeploymentSettings component, they aren't able to be saved.

## Expected behavior

Bounds can get saved.

In a new project, go to the deployment settings and try entering new bounds and saving. The bounds will disappear because they didn't get saved.

n/a

0f2a2b0d1ec75d4190e26195cb782312a1944853

evansiroky:
### Checklist

### Description

I found the placeholders describing bounds to be very hard to understand. This PR changes the wording of the placeholders to a hopefully much more easy-to-understand value. And then after that, I found that the bounding box field doesn't work in the DeploymentSettings component (#471). So I made another commit that fixes #471.

### DIFF
--- a/lib/manager/util/index.js
+++ b/lib/manager/util/index.js
@@ -297,32 +297,33 @@ export function parseBounds (val: string): {
     // Filter out any bad parsed values
     .filter(parsedValue => !isNaN(parsedValue))
 
-  if (bBox.length === 4) {
-    const [west, south, east, north] = bBox
-
-    // make sure all values are valid
-    validateVal(west, 'West', isValidLongitude)
-    validateVal(east, 'East', isValidLongitude)
-    validateVal(north, 'North', isValidLatitude)
-    validateVal(south, 'South', isValidLatitude)
-
-    if (errors.length > 0) {
-      return {
-        errors,
-        valid: false
-      }
-    }
-
-    // valid!
-    const bounds = {west, south, east, north}
+  if (bBox.length !== 4) {
     return {
-      bounds,
-      valid: true
+      errors: [`Incorrect amount of values. Found ${bBox.length}, expected 4.`],
+      valid: false
     }
   }
+
+  const [west, south, east, north] = bBox
+
+  // make sure all values are valid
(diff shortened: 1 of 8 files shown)


lib/manager/util/index.js
```diff
@@ -234,6 +240,92 @@ export function isEditingDisabled (
...
+    const bounds = {west, south, east, north}
+    return {
+      bounds,
+      valid: true
+    }
+  }
+  return {
+    errors: [`Incorrect amount of values. Found ${bBox.length}, expected 4.`],
+    valid: false
+  }
```
landonreed: This might be nitpicky, but I kind of prefer this invalid return statement to come before the above conditional block. Instead of using the `bBox.length === 4` condition to handle the valid number of args, changing it so that the incorrect number of args short circuits the validation check.

```
if (bBox.length !== 4) {
  return {
    errors: [`Incorrect amount of values. Found ${bBox.length}, expected 4.`],
    valid: false
  }
}

const [west, south, east, north] = bBox
// make sure all values are valid
validateVal(west, 'West', isValidLongitude)
validateVal(east, 'East', isValidLongitude)
validateVal(north, 'North', isValidLatitude)
validateVal(south, 'South', isValidLatitude)

if (errors.length > 0) {
  return {
    errors,
    valid: false
  }
}

// valid!
const bounds = {west, south, east, north}
return {
  bounds,
  valid: true
}
```

I can't really explain why that seems preferable to me...

evansiroky: It is actually a lot less indentation. Probably saves maybe 10 bytes in the file size.

evansiroky:
The error descriptions don't get shown anywhere yet. Certain form validation libraries ([like react-formal](http://jquense.github.io/react-formal/api/Field)) will list out the errors below the form inputs.

lib/manager/util/index.js
landonreed: #TrimTheFat

PR #505, opened 2019-10-21
Makes sure a user's email address is verified before watching a project/feed

### LINKED ISSUE #492: Resend Auth0 email verification link if unverified user clicks watch button
If a user clicks the "Watch" button to subscribe to notifications about a project, feed source, etc. and the user's email has not been verified (`user.profile.email_verified = false`), we should show a window.confirm with a message like (I'm open to wordsmithing): 

```
In order to receive email notification, you must first verify your email address. Would you like to resend the email verification message for your account?
```

A confirmation of this should trigger a request to datatools-server to use the Auth0 management API https://auth0.com/docs/api/management/v2#!/Jobs/post_verification_email to resend. 

It might also be worth adding some kind of warning style/color to the watch button if the email is not verified. Or maybe, instead we should add a resend button next to the email address on the under-utilized user profile page (along with a red notification dot on the 'My account' button on the home page).

image

evansiroky:
### Checklist

### Description

Fixes #492.

Requires companion PR in datatools-server: https://github.com/ibi-group/datatools-server/pull/254 for this PR to work.

### DIFF
--- a/i18n/english.yml
+++ b/i18n/english.yml
@@ -759,5 +759,8 @@ components:
     timestamp: File Timestamp
     version: version
   WatchButton:
+    emailVerificationConfirm: In order to receive email notification, you must first verify your email address. Would you like to resend the email verification message for your account?
     unwatch: Unwatch
+    verificationSendError: There was an error sending the email verification!
+    verificationSent: Email verification successfully sent! Please confirm check your email and confirm your email address. Then, you will need to refresh this page after waiting for about a minute for the change to take effect.
     watch: Watch
--- a/i18n/english.yml
+++ b/i18n/english.yml
@@ -762,5 +762,5 @@ components:
     emailVerificationConfirm: In order to receive email notification, you must first verify your email address. Would you like to resend the email verification message for your account?
     unwatch: Unwatch
     verificationSendError: There was an error sending the email verification!
-    verificationSent: Email verification successfully sent! Please confirm check your email and confirm your email address. Then, you will need to refresh this page after waiting for about a minute for the change to take effect.
+    verificationSent: Please check your inbox and confirm your email address by clicking the verify link. You will then need to refresh this page after a moment or two and re-click the Watch button to subscribe to notifications.
     watch: Watch
--- a/lib/common/containers/WatchButton.js
+++ b/lib/common/containers/WatchButton.js
@@ -4,7 +4,11 @@ import React, {Component} from 'react'
 import {Button, Glyphicon, MenuItem} from 'react-bootstrap'
 import {connect} from 'react-redux'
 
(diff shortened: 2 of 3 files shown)


lib/common/containers/WatchButton.js
```diff
@@ -18,16 +22,58 @@ type ContainerProps = {
...
+      subscriptionType,
+      target,
+      updateTarget,
+      user
+    } = this.props
+    if (user.profile) {
+      if (!isWatching && !user.profile.email_verified) {
+        if (confirm(this.messages('emailVerificationConfirm'))) {
+          resendEmailConfirmation(user)
+            .then((result: boolean) => {
```
landonreed: I was under the impression that you were somewhat opposed to using thunk in this way (chaining actions like this). The other way to handle this would be to just interact with the redux store. I don't have hugely strong feelings about this, but was under the impression that this was the direction we were heading.

EDIT: On second thought, this approach is growing on me -- it's so ephemeral that putting the result in the store seems silly and I like that you're using `setErrorMessage` in the component so that we can set the message based on `WatchButton` messages.

i18n/english.yml
```diff
@@ -759,5 +759,8 @@ components:
     timestamp: File Timestamp
     version: version
   WatchButton:
+    emailVerificationConfirm: In order to receive email notification, you must first verify your email address. Would you like to resend the email verification message for your account?
     unwatch: Unwatch
+    verificationSendError: There was an error sending the email verification!
+    verificationSent: Email verification successfully sent! Please confirm check your email and confirm your email address. Then, you will need to refresh this page after waiting for about a minute for the change to take effect.
```
landonreed: `verificationSent` typo:

Recommend changing to:

```
Please check your inbox and confirm your email address by clicking the verify link. You will then need to refresh this page after a moment or two and re-click the Watch button to subscribe to notifications.
```

lib/manager/actions/user.js
```diff
@@ -127,6 +127,24 @@ export function logout (logoutFromAuth0: ?boolean = true) {
   return userLogout()
 }
 
+/**
+ * Resends the email confirmation for a given user. Returns a promise with a
+ * resulting value of true if the email was successfully sent.
```
landonreed: Maybe change first sentence to:
```
Resends the Auth0 email confirmation message (which is auto-sent on user signup) for a given user.
```

lib/common/containers/WatchButton.js
evansiroky: Yeah, it sorta seemed easier to do it this way since messages apply only to components, otherwise this logic would all be in an action.

evansiroky:
This change won't affect the e2e tests because there's no e2e test for watching/unwatching a feed/project.

PR #527, opened 2020-01-26
Validation issue prioritization

### LINKED ISSUE #477: Indicate validation issue priority in validation table
In the list of validation issues for a given feed version, there is no indication of priority given to the various issue types. In other words, there is no way to distinguish whether an issue could be considered an error vs. a warning -- e.g., something that would break feed ingestion by a third party service vs. just affect how a name field might be rendered.

We already have a measure of priority stored in the gtfs-lib class that deals with validation issues. We just need to display this in the UI using either the binary error vs. warning classification or the multi-tiered priority classification.

evansiroky:
### Checklist

### Description

This PR seeks to fix #477 by modifying the version validation error viewer in two ways:
- it will sort errors based off of priority level and then alphabetically
- it shows different icons based on error severity

This PR depends on the https://github.com/ibi-group/datatools-server/pull/275 PR to datatools-server that uses the version of gtfs-lib that includes the graphQL changes.

### DIFF
--- a/i18n/english.yml
+++ b/i18n/english.yml
@@ -382,7 +382,18 @@ components:
       stops: Stop issues
       trips: Trip issues
     noResults: No validation results to show.
-    title: Validation Results
+    tips:
+      DATE_NO_SERVICE: "If the transit service does not operate on weekends,
+      some or all of these validation issue may be ignored. Similarly, holidays
+      for which there is no transit service running may appear in this list."
+      FEED_TRAVEL_TIMES_ROUNDED: "This is a common feature of GTFS feeds that do
+      not use down-to-the-second precision for arrival/departure times. However,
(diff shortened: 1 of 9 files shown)


binh-dam-ibigroup:
The way it is depicted in Landon's post might suggest there could be, within one category, there might be critical issues and less critical issues (e.g. 50 critical unused stops and 214 less critical stops).
I would place the icon to the left of the text to avoid that confusion (keep the count to the right).

landonreed:
@binh-dam-ibigroup, I don't 100% follow the comment, but here's what it looks like with the icons on the right. I think it does look better, but I'm not crazy about how this looks. I'm wondering if there's a better way to format all of this.

image

evansiroky:
Other PRs have been merged in. Good to go, @landonreed?

PR #529, opened 2020-01-30
refactor(validation): update UI for issue type/count display

landonreed:
### Checklist

### Description

Refactor the display of validation issues/types/counts:
image

### DIFF
--- a/lib/style.css
+++ b/lib/style.css
@@ -179,20 +179,72 @@ td.feed-source-info {
 
 /* feed validation error types */
 
-.gtfs-error-high {
+.gtfs-error-HIGH-label {
+  color: #c9302c;
+  background-color: #ffffff;
+  border: 1px #c9302c solid;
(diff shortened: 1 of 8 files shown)


lib/style.css
```diff
@@ -179,20 +179,72 @@ td.feed-source-info {
 
 /* feed validation error types */
 
-.gtfs-error-high {
+.gtfs-error-HIGH-label {
```
binh-dam-ibigroup: Are these errors to circumvent the limited colors available out of the box in the version of react-bootstrap we are using?

landonreed: Yea, I believe that's the intent. There may be some disagreement between the react-bs colors and these though, so we might want to reconsider the mixing of the bs label colors with these new ones. I do like the pattern of using the priority id in the class name selector rather than using a conditional to select the color within the code.

Issue #544, opened 2020-03-16
MTC Enhancements 03/2020


landonreed:
- [x] Update feed version dropdown (reverse sort, retrieval method filtering)
- [x] Add field length validation checks (per requirements set forth in [511 Transit_Data Guidelines_V2.0](https://github.com/ibi-group/datatools-ui/files/4438625/511.Transit_Data.Guidelines_V2.0_3-27-2020.pdf))
- [x] Add ability to transform GTFS feeds, including:
  - [x] Replace file (e.g., `stop_attributes.txt`) in new version with file from previous version)
  - [x] Replace file in new version with user-defined CSV data

binh-dam-ibigroup:
Closing this issue, the transformations were handled by #579, field length validation in #648.

PR #545, opened 2020-03-16
MTC Enhancements March 2020

landonreed:
### Checklist

### Description

MTC enhancements re #544. 
- Reverse version dropdown sort
- Do not allow selecting of merged feeds in merge feeds dropdown
- Add retrieval method icon to dropdown items/version summary


lib/manager/components/version/FeedVersionDetails.js
```diff
@@ -51,12 +52,30 @@ export default class FeedVersionDetails extends Component<Props> {
...
   }
 
+  _mergeItemFormatter = (v: FeedVersion, i: number, activeVersion: FeedVersion) => {
+    const name = v.id === activeVersion.id
+      ? '(Cannot merge with self)'
+      : v.retrievalMethod === 'SERVICE_PERIOD_MERGE'
+        ? '(Cannot re-merge feed)'
+        : v.name
+    const disabled = v.retrievalMethod === 'SERVICE_PERIOD_MERGE' ||
+      v.id === activeVersion.id
```
evansiroky: Nitpick: might be better to use if/else if blocks here to make things more readable and avoid duplicate code.

lib/manager/components/version/VersionRetrievalBadge.js
```diff
@@ -0,0 +1,40 @@
+// @flow
+
+import Icon from '@conveyal/woonerf/components/icon'
+import * as React from 'react'
```
evansiroky: I don't think we need to import everything from React here. Seems like `React, {Component}` should suffice.

lib/manager/components/version/VersionRetrievalBadge.js
```diff
@@ -0,0 +1,40 @@
...
+    case 'MANUALLY_UPLOADED':
+      return 'upload'
+    case 'FETCHED_AUTOMATICALLY':
+      return 'cloud-download'
+    default:
+      return 'file-archive-o'
+  }
+}
+
+export default class VersionRetrievalBadge extends React.Component<Props> {
```
evansiroky: This could be just a function, no need to make it a full component.

lib/manager/components/version/VersionSelectorDropdown.js
```diff
@@ -0,0 +1,77 @@
...
+          }}
+          type='check' />
+        : null}
+      {version.version}. {version.name}{' '}
+      <VersionRetrievalBadge version={version} />
+    </MenuItem>
+  )
+}
+
+export default class VersionSelectorDropdown extends React.Component<Props> {
```
evansiroky: This could be just a function, no need to make it a full component.

lib/manager/components/version/VersionSelectorDropdown.js
```diff
@@ -0,0 +1,77 @@
...
+type Props = {
+  dropdownProps: any,
+  header?: string,
+  itemFormatter: (FeedVersion, number, FeedVersion) => React.Node,
+  title: React.Node,
+  version: FeedVersion,
+  versions: ?Array<FeedVersion>
+}
+
+const defaultItemFormatter = (version: FeedVersion, index: number, activeVersion: FeedVersion) => {
```
evansiroky: Make this into a functional component. Also, is the index argument needed/used?

lib/manager/actions/status.js
```diff
@@ -134,6 +136,51 @@ export async function fetchAppInfo () {
...
+      const tableAndId = split.splice(1, 1)
+      remappedIdStrings.push(`${tableAndId.join(':')} -> ${result.remappedIds[key]}`)
+    }
+    details.push('Remapped IDs: ' + remappedIdStrings.join(', '))
+  }
+  if (result.skippedIds.length > 0) {
+    const skippedRecordsForTables = {}
+    result.skippedIds.forEach(id => {
+      const table = id.split(':')[0]
+      if (skippedRecordsForTables[table]) {
```
binh-dam-ibigroup: Suggestion: 
`const skippedRecords = skippedRecordsForTables[table]
skippedRecordsForTables[table] = skippedRecords ? 1 : skippedRecords + 1 
`

lib/manager/actions/status.js
```diff
@@ -134,6 +136,51 @@ export async function fetchAppInfo () {
...
+      ? 'Warning: Errors encountered during feed merge!'
+      : 'Feed merge was successful!',
+    body: result.failed
+      ? `Merge failed with ${result.errorCount} errors. ${result.failureReasons.join(', ')}`
+      : `Merge was completed successfully. A new version will be processed/validated containing the resulting feed.`,
+    detail: details.join('\n')
+  }
+}
+
 /* eslint-disable complexity */
```
binh-dam-ibigroup: Can you remove `/* eslint-disable complexity */` now?

lib/manager/actions/status.js
```diff
@@ -203,60 +250,21 @@ export function handleFinishedJob (job: ServerJob) {
...
-            dispatch(setStatusModal({
-              title: result.failed
-                ? 'Warning: Errors encountered during feed merge!'
-                : 'Feed merge was successful!',
-              body: result.failed
-                ? `Merge failed with ${result.errorCount} errors. ${result.failureReasons.join(', ')}`
-                : `Merge was completed successfully. A new version will be processed/validated containing the resulting feed.`,
-              detail: details.join('\n')
-            }))
+            const modalContent = getMergeFeedModalContent(result)
```
binh-dam-ibigroup: Inline this.

lib/manager/components/version/VersionRetrievalBadge.js
landonreed: I think that was leftover from when I was trying to specify a flow type of React.Node.

lib/manager/components/version/FeedVersionDetails.js
landonreed: Something like:
```
let name = v.name
    let disabled = false
    if (v.id === activeVersion.id) {
      name = '(Cannot merge with self)'
      disabled = true
    }
    if (v.retrievalMethod === 'SERVICE_PERIOD_MERGE') {
      name = '(Cannot re-merge feed)'
      disabled = true
    }
```
?

lib/manager/components/version/VersionSelectorDropdown.js
```diff
@@ -0,0 +1,77 @@
...
+    itemFormatter: defaultItemFormatter
+  }
+
+  render () {
+    const {
+      dropdownProps,
+      header,
+      itemFormatter,
+      title,
+      version,
```
binh-dam-ibigroup: Suggestion: also rename `version` to `activeVersion`.

lib/manager/components/version/FeedVersionDetails.js
binh-dam-ibigroup: Actually does it make sense at all to display versions that cannot be merged?

lib/manager/actions/status.js
landonreed: I'm not sure if this is needed anywhere anymore. When I remove these `yarn lint` runs fine. @evansiroky can you weigh in on whether these are still needed?

lib/manager/components/version/FeedVersionDetails.js
landonreed: I think it works either way. I'm inclined to just disable them so they don't appear to be missing to the user without explanation.

lib/manager/actions/status.js
landonreed: I think your ternary is backwards?

binh-dam-ibigroup: Probably 🤔

landonreed: How about:
```
skippedRecordsForTables[table] = (skippedRecordsForTables[table] || 0) + 1
```
That doesn't seem too clever, but is nice and concise.

binh-dam-ibigroup: That works too.

lib/manager/components/version/VersionSelectorDropdown.js
landonreed: Leaving it named version conforms with the name of the prop in other version components. I do somewhat agree with you on this item, but I think it should be handled in a separate PR and deserves a little more consideration on whether we should make the change.

landonreed:
@evansiroky, I have addressed your comments. @binh-dam-ibigroup, I addressed one comment, but skipped the eslint disable one based on our discussion on IBI chat.  Also, I'm not keen on changing the prop name unless we did it wholesale throughout other components, which seems a bit out of scope.

PR #551, opened 2020-04-06
fix(gtfs.yml): add more currency types for fares

### LINKED ISSUE #550: Add more currencies for fares
Data Tools UI currently only supports a handful of [currency types](https://en.wikipedia.org/wiki/ISO_4217#Active_codes) (USD, EUR, and GBP). We should add more to cover more transit markets (especially North America).

landonreed:
### Checklist

### Description

fixes #550 by adding top currencies traded in world (and major North American currency types).

### DIFF
--- a/gtfs.yml
+++ b/gtfs.yml
@@ -629,9 +629,28 @@
     inputType: DROPDOWN
     bulkEditEnabled: true
     options:
-    - value: 'USD'
-    - value: 'EUR'
-    - value: 'GBP'
+    - value: USD
+      text: US dollar (USD)
+    - value: AUD
+      text: Australian dollar (AUD)
+    - value: CAD
+      text: Canadian dollar (CAD)
+    - value: CHF
+      text: Swiss franc (CHF)
+    - value: CNH
+      text: Chinese renminbi (CNH)
+    - value: EUR
+      text: Euro (EUR)
+    - value: GBP
+      text: Pound sterling (GBP)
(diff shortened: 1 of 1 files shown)


landonreed:
Yea, perhaps you're right @evansiroky.

binh-dam-ibigroup:
Instead of keying them in manually, can that be retrieved from some other source?

landonreed:
Actually, I'm not really sure including all of them is warranted @evansiroky and @binh-dam-ibigroup. Having hundreds of values really just clutters the dropdown immensely. Sure, specifying some obscure currency is valid, but if our only users of the system are in North America, does it really help to have "Platinum" as a currency type in here?

binh-dam-ibigroup:
I agree with @landonreed, it's fine to key in a few like in this PR, but it doesn't make sense to key in everything, especially if we just copy from some other place.

evansiroky:
I'm still a fan of allowing all possible values. The [GTFS spec is clear](http://gtfs.org/reference/static/#field-types) that any ISO 4217 currency code is a valid value. Since we're making software that produces GTFS feeds I don't think it makes sense to exclude valid possible values. Go ahead and overrule me if you like though.

landonreed:
@evansiroky, how do you propose we hide the values with react-select? I'm open to the idea. Feel free to jump on https://github.com/ibi-group/datatools-ui/pull/566/files.

PR #573, opened 2020-05-04
Fix pattern name editor bug

### LINKED ISSUE #556: GTFS Editor: Pattern Name Edit Reverts User Changes Due to Editor Lock Request
While editing a trip pattern name in the GTFS Editor, if the EditableTextField is in an edit state while the editor makes a lock request (the editor occasionally makes these to keep a session lock on the feed being edited to prevent multiple editing sessions), the changes to the pattern name will be reverted back to the previous value.

To recreate:
1. Open a feed in the editor.
2. Select a route, and create a new pattern.
3. Click edit name, quickly change the value (without saving) and then wait a few seconds.

landonreed:
### Checklist

### Description

fix #556 by not resetting EditableTextField value if editing session is in progress.

### DIFF
--- a/lib/common/components/EditableTextField.js
+++ b/lib/common/components/EditableTextField.js
@@ -48,20 +48,21 @@ export default class EditableTextField extends Component<Props, State> {
 
   componentWillReceiveProps (nextProps: Props) {
     const {initialValue, isEditing, value} = this.state
-    const valueChanged = this.props.value !== nextProps.value
-    if (value !== nextProps.value && valueChanged && !isEditing) {
+    const {value: nextValue} = nextProps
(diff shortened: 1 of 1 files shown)


lib/common/components/EditableTextField.js
```diff
@@ -47,12 +47,20 @@ export default class EditableTextField extends Component<Props, State> {
   }
 
   componentWillReceiveProps (nextProps: Props) {
-    if (this.state.value !== nextProps.value) {
-      // Update value if externally changed.
+    const {initialValue, isEditing, value} = this.state
```
binh-dam-ibigroup, line 50: Also extract nextProps.value into its own variable.

PR #588, opened 2020-06-26
Add trip short name

### LINKED ISSUE #569: Timetable Editor: Add trip_short_name field
This field already exists in the database for trips, we just need a place to edit it in the timetable editor. Long term, we should probably make this a part of the trip pattern, but that doesn't currently exist for trip_headsign either, so an incremental fix is just to update the editor to include this field.

Starting point for this would be to add an entry for `tripShortname` below `tripHeadsign` here:

https://github.com/ibi-group/datatools-ui/blob/b3f79afeff48148ca8faedc7f33f04ac2af9c37d/lib/editor/selectors/timetable.js#L46-L52

philip-cline:
### Checklist

- [ ✓] Appropriate branch selected _(all PRs must first be merged to `dev` before they can be merged to `master`)_
- [✓ ] Any modified or new methods or classes have helpful JSDoc and code is thoroughly commented
- [ ✓] The description lists all applicable issues this PR seeks to resolve
- [ ✓] The description lists any configuration setting(s) that differ from the default settings
- [ ✓] All tests and CI builds passing

### Description

Added a trip_short_name field (Closes #569 ) to the timetable editor (by adding object in timetable.js). Also added tests to end-to-end.js to enter a value in trip_short_name.

### DIFF
--- a/lib/editor/selectors/timetable.js
+++ b/lib/editor/selectors/timetable.js
@@ -52,7 +52,7 @@ export const getTimetableColumns = createSelector(
       },
       {
         name: 'Trip Short Name',
-        width: 200,
+        width: 120,
         key: 'tripShortName',
         type: 'TEXT',
         placeholder: '801'
--- a/lib/editor/selectors/timetable.js
+++ b/lib/editor/selectors/timetable.js
@@ -49,6 +49,13 @@ export const getTimetableColumns = createSelector(
         key: 'tripHeadsign',
         type: 'TEXT',
         placeholder: 'Destination via Transfer Center'
+      },
+      {
+        name: 'Trip Short Name',
+        width: 200,
+        key: 'tripShortName',
+        type: 'TEXT',
+        placeholder: '801'
       }
     ]
     if (pattern && pattern.patternStops) {
--- a/__tests__/end-to-end.js
+++ b/__tests__/end-to-end.js
@@ -2476,6 +2476,7 @@ describe('end-to-end', () => {
           await page.keyboard.press('Tab')
           await page.keyboard.press('Enter') 
 
+
           // Laurel Dr arrival
           await page.keyboard.type('1234')
           await page.keyboard.press('Tab')
--- a/__tests__/end-to-end.js
+++ b/__tests__/end-to-end.js
@@ -2471,7 +2471,7 @@ describe('end-to-end', () => {
           await page.keyboard.press('Tab')
           await page.keyboard.press('Enter')
 
-          // trip short name 
+          // trip short name
           await page.keyboard.type('test-trip-short-name')
           await page.keyboard.press('Tab')
           await page.keyboard.press('Enter')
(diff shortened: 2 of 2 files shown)


lib/editor/selectors/timetable.js
```diff
@@ -49,6 +49,13 @@ export const getTimetableColumns = createSelector(
         key: 'tripHeadsign',
         type: 'TEXT',
         placeholder: 'Destination via Transfer Center'
+      },
+      {
+        name: 'Trip Short Name',
+        width: 200,
+        key: 'tripShortName',
+        type: 'TEXT',
+        placeholder: '801'
```
binh-dam-ibigroup, line 58: How did you come up with the value for this?

philip-cline:
Re @binh-dam-ibigroup's comments: I found the trip_short_name field at this example: https://support.google.com/transitpartners/answer/6377422?hl=en. 

The GTFS spec outlines trip_short_name as "Public facing text used to identify the trip to riders, for instance, to identify train numbers for commuter rail trips. If riders do not commonly rely on trip names, leave this field empty. A trip_short_name value, if provided, should uniquely identify a trip within a service day; it should not be used for destination names or limited/express designations." 

As @landonreed said, it's mostly reserved for vehicle numbers so perhaps '801' doesn't represent a great example. The MBTA uses three digits such as '005' for their commuter rail trip_short_names - perhaps this would be a better example to use. Let me know and I will update.

lib/editor/selectors/timetable.js
binh-dam-ibigroup: In light of your comment @philip-cline, my preference would be something like `Train 012 (or, just 012)`, or no placeholder at all. I'll let @landonreed break the tie.

landonreed: I think `801` is fine. I'm not really sure how that's any different from something like `012`. If we wanted to modify all of the placeholders to do something different (besides offering a single example value) that's a separate concern.

binh-dam-ibigroup: Sure

philip-cline:
@landonreed, I updated the width to be 120 for this and kept the 801 placeholder.

evansiroky:
> ### Checklist
> * [ ✓] Appropriate branch selected _(all PRs must first be merged to `dev` before they can be merged to `master`)_
> * [✓ ] Any modified or new methods or classes have helpful JSDoc and code is thoroughly commented
> * [ ✓] The description lists all applicable issues this PR seeks to resolve
> * [ ✓] The description lists any configuration setting(s) that differ from the default settings
> * [ ✓] All tests and CI builds passing
> 
> ### Description
> Added a trip_short_name field (Closes #569 ) to the timetable editor (by adding object in timetable.js). Also added tests to end-to-end.js to enter a value in trip_short_name.

@philip-cline FYI, the markdown way to add checks is to write it as `[x]`.

Example:

- [x] checked
- [ ] unchecked

PR #613, opened 2020-09-22
Fix e2e tests again

evansiroky:
### Checklist

- [x] e2e tests are all passing See https://travis-ci.org/github/ibi-group/datatools-ui/builds/729458864

### Description

This PR extends the amount of time (from 2s to 30s) to wait after creating a new Auth0 user before checking if it now is returned from the Auth0 API. Prior runs of the e2e tests were failing because the original wait time was not long enough.

### DIFF
--- a/__tests__/end-to-end.js
+++ b/__tests__/end-to-end.js
@@ -885,7 +885,7 @@ describe('end-to-end', () => {
       await click('[data-test-id="confirm-create-user-button"]')
 
       // wait for user to be saved
-      await wait(2000, 'for user to be created')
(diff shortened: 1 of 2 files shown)


landonreed (approved):
Seems like a reasonable enough fix. Is 30 seconds too long though?

evansiroky:
> Seems like a reasonable enough fix. Is 30 seconds too long though?

It's long enough to probably handle any lag on Auth0's end, but short enough that the overall test timeout still probably won't be met.

binh-dam-ibigroup (approved):

PR #615, opened 2020-10-03
Lazy-load server data with ec2 instance info

evansiroky:
### Checklist

### Description

This PR changes the way data is loaded for the ServerSettings admin component. Whenever a panel showing the details about a server is expanded, the UI will now make a call to a new endpoint in datatools-server to load the full information about the server which could include data about ec2 instances. This is done to avoid making unnecessary calls to AWS until that data is actually needed.

This PR requires using the code in https://github.com/ibi-group/datatools-server/pull/343.

### DIFF
--- a/lib/admin/components/ServerSettings.js
+++ b/lib/admin/components/ServerSettings.js
@@ -36,6 +36,7 @@ import type { AppState, AdminServersState } from '../../types/reducers'
 type ContainerProps = {editDisabled: boolean}
 type Props = ContainerProps & {
   deleteServer: typeof adminActions.deleteServer,
+  fetchServer: typeof adminActions.fetchServer,
   fetchServers: typeof adminActions.fetchServers,
   otpServers: AdminServersState,
   projects: Array<Project>,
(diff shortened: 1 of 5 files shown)


lib/admin/reducers/servers.js
```diff
@@ -19,6 +19,18 @@ const servers = (state: AdminServersState = defaultState, action: Action): Admin
...
       })
+    case 'RECEIVE_SERVER':
+      const serverData = action.payload
+      if (state.data) {
+        const serverIdx = state.data.findIndex(
+          server => server.id === serverData.id
+        )
+        return update(state, {
+          isFetching: { $set: false },
+          data: { [serverIdx]: { $set: action.payload } }
```
landonreed, line 30: Not necessarily needed, but seems like an opportunity to reuse the serverData variable?

lib/admin/components/ServerSettings.js
```diff
@@ -149,6 +150,10 @@ class ServerSettings extends Component<Props, State> {
     }
   }
 
+  _onExpandServer = (index: number) => {
+    this.props.fetchServer(this.state.otpServers[index].id)
+  }
+
   _onSaveServer = (index: number) => this.props.updateServer(this.state.otpServers[index])
```
landonreed, line 157: What happens if a server is expanded, you apply updates, and click save? Are the server ec2 instances no longer shown?

landonreed (changes requested):
One minor comment, one that probably needs addressing.

binh-dam-ibigroup (approved):
Subject to @landonreed's comment about refetching EC2 instances on an expanded server panel.

evansiroky:
Good catch. I just pushed a change that addresses this.

landonreed (approved):

PR #528, opened 2020-01-28
Indicate summary changes between feed versions

binh-dam-ibigroup:
### Checklist

- [X] Appropriate branch selected _(all PRs must first be merged to `dev` before they can be merged to `master`)_
- [ ] Any modified or new methods or classes have helpful JSDoc and code is thoroughly commented
- [X] The description lists all applicable issues this PR seeks to resolve
- NA - The description lists any configuration setting(s) that differ from the default settings
- [ ] All tests and CI builds passing

### Description

This fixes #384. With this patch, when a user selects a feed version, increments are shown for counts of validation issues, agencies, routes, stops, trips, stop times compared to the immediate previous version. The user can also compare with other previous versions as well.

image

PR #634, opened 2020-12-08
Feed diffs ltr

landonreed:
Refactors things a bit for #528

### DIFF
--- a/lib/manager/components/version/FeedVersionTabs.js
+++ b/lib/manager/components/version/FeedVersionTabs.js
@@ -2,18 +2,19 @@
 
 import moment from 'moment'
 import React, {Component} from 'react'
-import {Col, Label as BsLabel, ListGroupItem, Row, Tab, Tabs} from 'react-bootstrap'
+import {Col, ListGroupItem, Row, Tab, Tabs} from 'react-bootstrap'
 import numeral from 'numeral'
 
 import {getComponentMessages, isModuleEnabled} from '../../../common/util/config'
+import DeltaStat from './DeltaStat'
 import Patterns from '../reporter/containers/Patterns'
 import Routes from '../reporter/containers/Routes'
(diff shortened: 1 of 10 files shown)


lib/manager/components/version/FeedVersionTabs.js
```diff
@@ -132,6 +146,16 @@ export default class FeedVersionTabs extends Component<Props> {
                 <p style={{marginBottom: '0px'}}>{this.messages('daysActive')}</p>
               </Col>
             </Row>
+            <Row style={{marginTop: '40px'}}>
```
binh-dam-ibigroup, line 149: @landonreed what do you think of the below:
- With the proposed positioning of the selector for the compared version, I still think the best place for the validity date comparator is below the existing valid date heading.
- One alternate layout could be to move the compared version selector inside the top portion of the summary tab, in which case it makes sense to keep the date comparison chart at this location.

binh-dam-ibigroup (commented):
See my comment at https://github.com/ibi-group/datatools-ui/pull/634/files#r550260760. I am otherwise inclined to accept your changes.

lib/manager/components/version/FeedVersionTabs.js
landonreed: My main issue with placing the chart in the valid date heading: when there's a GTFS+ blocking issue, the warning text gets displaced by the chart (see screenshot). Also, I think it makes some sense to place the chart with the other "date oriented" charts (i.e., the histograms showing trips/service hours).

image

Given that the compared version selector does affect elements across the entire page, maybe it makes sense to move it next to the green "Create new version button". There is some other work on the roadmap to move the "Edit feed" to its own box below the page selector (Version summary, validation issues, etc.), so that could work well.

binh-dam-ibigroup (approved):

PR #638, opened 2021-02-02
Add OTP 2 deployment enhancements (OTP 2 checkbox and custom files)

### LINKED ISSUE #519: Remove references to R5
## Observed behavior

Deployment in Data Tools offers an R5 option for a tool no longer used for OTP.

## Expected behavior

Option, and references in code, should be removed.

evansiroky:
### Checklist

### Description

This adds <strike>a checkbox</strike> a dropdown to the DeploymentViewer component so that a user can select whether a jar file is an OTP 2.x jar. 

Also, it is now possible to add custom files to a deployment. These show up beneath the build and router config items.

This works in conjunction with https://github.com/ibi-group/datatools-server/pull/363.

This PR also removes references to r5 (fixes #519)


binh-dam-ibigroup (approved):
I think this is good for the time being.

evansiroky:
Assigning back to myself to make deployments be able to include extra files (otp-config.json and walk-config.json).

evansiroky:
Ready for a review again with custom file support for deployments.

lib/manager/components/DeploymentViewer.js
```diff
@@ -665,3 +748,205 @@ class CustomConfig extends Component<{
...
+              {!customFile.uri && (
+                <HelpBlock>Enter either a HTTP(S) URL or AWS S3 URI</HelpBlock>
+              )}
+            </span>
+          )}
+        </FormGroup>
+      </div>
+    )
+  }
+}
```
landonreed, line 952: I think it's safe to say that this file is getting far too large! Could you please split it up a bit?

binh-dam-ibigroup (changes requested):
Please address removing blank custom file entries that the user does not save, and the X in the OTP version dropdowns.

landonreed (changes requested):
@evansiroky, I have more to look at, but before you do any refactoring on the custom file stuff, I think we need to come up with a more standard interface for setting these files that's shared between build/router config and the custom files. Perhaps I'll suggest some changes in a PR.

landonreed:
Marking as Blocked until #641 is merged.

evansiroky:
I added some more validation checks for custom files.

lib/manager/components/DeploymentViewer.js
```diff
@@ -407,31 +440,35 @@ export default class DeploymentViewer extends Component<Props, State> {
...
-              </Checkbox>
-            }
+              value={deployment.tripPlannerVersion}
+              onChange={this._onUpdateTripPlannerVersion}
+              options={TRIP_PLANNER_VERSIONS}
+            />
+            {isOtp && (
+              <div>
+                OTP jar file
+                <Select
```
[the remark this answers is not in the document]
binh-dam-ibigroup: I'll give a pass on this one, I understand now why`<Select>` was preferred over `<FomrControl>`.

lib/manager/components/deployment/CustomFileEditor.js
```diff
@@ -0,0 +1,299 @@
...
+    const either = !isEditing && !canEdit
+    return (
+      <ButtonToolbar>
+        <Button
+          bsSize='xsmall'
+          disabled={either}
+          onClick={canEdit ? this._onEdit : this._onCancelEditing}
+        >
+          {isEditing ? 'Cancel' : 'Edit'}
+        </Button>
```
binh-dam-ibigroup, line 176: See my other comment regarding hiding Cancel when creating a new entry.

lib/manager/components/deployment/DeploymentConfigurationsPanel.js
```diff
@@ -0,0 +1,280 @@
...
+
+  _updateDeployment = (props: {[string]: any}) => {
+    const {deployment, updateDeployment} = this.props
+    updateDeployment(deployment, props)
+  }
+
+  render () {
+    const { deployment, updateDeployment } = this.props
+    const { customFileEditIdx, otp: options } = this.state
+    const isOtp = deployment.tripPlannerVersion.startsWith('OTP_')
```
binh-dam-ibigroup, line 161: Is this always true?

binh-dam-ibigroup (approved):
LGTM. Please consider the minor suggestions regarding hiding the Cancel button when creating a new custom file and one maybe obsolete variable.

lib/manager/components/deployment/CustomFileEditor.js
```diff
@@ -40,11 +40,37 @@ export default class CustomFileEditor extends Component<Props, {
...
+    return (fileSource === 'uri' && !!customFile.uri) ||
+      (!!customFile.contents && !!customFile.filename)
+  }
+
+  /**
+   * Makes sure that a uri or some raw input has been added
+   */
+  _hasContents = (): boolean => {
+    const {model: customFile} = this.state
+    return !!customFile.contents || !!customFile.uri
```
landonreed, line 58: Here and above in `_fileNameValid` I would suggest using `Boolean(customFile.contents)` rather than `!!customFile.contents`. I just find it infinitely more readable.

lib/manager/components/deployment/CustomFileEditor.js
```diff
@@ -40,11 +40,37 @@ export default class CustomFileEditor extends Component<Props, {
...
 
-  _isEditing= () => this.props.idx === this.props.customFileEditIdx
+  /**
+   * Makes sure that the filename is valid. The filename is optional when a uri
+   * is provided, but is required when entering raw input.
+   */
+  _fileNameValid = (): boolean => {
+    const {fileSource, model: customFile} = this.state
+    return (fileSource === 'uri' && !!customFile.uri) ||
+      (!!customFile.contents && !!customFile.filename)
```
landonreed, line 50: Does this check for URI belong here? It seems like a simpler approach would just be:
```suggestion
    const {contents, filename} = customFile
    return fileSource === 'raw' && Boolean(contents) && Boolean(filename)
```

landonreed (approved):
Looks good, just a couple of suggestions.

lib/manager/components/deployment/CustomFileEditor.js
evansiroky: Yes, it's required. The proposed change would return `false` for any custom file with a URI which is not correct since a filename is optional when a URI is provided.

evansiroky: After looking at this more closely I ended up rewriting this so that it always returns `true` for a custom file with a URI. See 00445ee33e7b02c77a650d4a4098e64b6c528c67

lib/manager/components/deployment/CustomFileEditor.js
evansiroky: I don't think `!!` is unreadable, but whatever, I made the change in 00445ee33e7b02c77a650d4a4098e64b6c528c67

lib/manager/components/deployment/DeploymentConfigurationsPanel.js
evansiroky: For now, yes

lib/manager/components/deployment/CustomFileEditor.js
```diff
@@ -0,0 +1,299 @@
...
+    this.setState({
+      model: {
+        ...this.state.model,
+        uri: evt.target.value
+      }
+    })
+  }
+
+  _onCancelEditing = () => {
+    const { customFile, onCancelEditing } = this.props
```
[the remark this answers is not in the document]
evansiroky: Cancel button hiding logic implemented in 00445ee33e7b02c77a650d4a4098e64b6c528c67

lib/manager/components/deployment/CustomFileEditor.js
evansiroky: Cancel button hiding logic implemented in 00445ee

evansiroky:
Assigning back for one last review.

lib/manager/components/deployment/DeploymentConfigurationsPanel.js
```diff
@@ -31,6 +31,16 @@ const TRIP_PLANNER_VERSIONS = [
   { label: 'OTP 2.X', value: 'OTP_2' }
 ]
 
+function calculateCustomFileKey (customFile: CustomFile, idx: number): string {
+  return `${
+    customFile.filename || ''
+  }-${
+    customFile.contents || ''
```
binh-dam-ibigroup, line 38: Would a timestamp or a uuid better than the file contents for identification purposes? (If it's a uuid, we won't need the other attributes to generate a unique key)

binh-dam-ibigroup (approved):
Please see the incidental comments. Also, please add in the PR description that it fixes #519.

lib/manager/components/deployment/DeploymentConfigurationsPanel.js
evansiroky: This is the tricky thing about the custom files since they're saved as an array within the deployment object. Perhaps they should be in their own table and then they would have IDs. But since they don't, I think it makes sense to use these items to create a composite key.

lib/manager/components/deployment/DeploymentConfigurationsPanel.js
```diff
@@ -31,6 +31,16 @@ const TRIP_PLANNER_VERSIONS = [
...
+  }-${
+    customFile.uri || ''
+  }-${idx}`
```
landonreed, line 41: This formatting is kind of painful to me. OK if we "one-line" this if we just destructure the various `customFile` fields?

```suggestion
  const {contents, filename, uri} = customFile
  return `${filename || ''}-${contents || ''}-${uri || ''}-${idx}`
```

landonreed (approved):
Looks good, just one minor comment/suggestion.

PR #640, opened 2021-02-11
Replace Travis CI with GitHub Actions

landonreed:
### Checklist

### Description

Per our CI strategy with other repos, this PR replaces Travis CI with GitHub Actions. E2E is currently disabled.


.github/workflows/node-ci.yml
```diff
@@ -0,0 +1,114 @@
+name: Node.js CI
+
+on: [push, pull_request]
```
binh-dam-ibigroup, line 3: Do we need the tests to run on pushes prior to a pull request being made? (This is to avoid running tests twice unnecessarily)

binh-dam-ibigroup (approved):
Looks good, please advise regarding running the test script twice.

.github/workflows/node-ci.yml
landonreed: I'm not too bothered by this. I think it's good to know if a commit passes CI even if it doesn't exist within a pull request.

evansiroky (changes requested):
Can you push at least one commit that hardcode-enables the E2E tests so that it can be verified that the E2E tests can actually start? It's fine if they don't pass completely, but I do want to see them actually get started in at least one commit workflow run.

landonreed:
Good thought, @evansiroky. Here's evidence that e2e will run (and fail): https://github.com/ibi-group/datatools-ui/pull/640/checks?check_run_id=1927768338

evansiroky (changes requested):
> Here's evidence that e2e will run (and fail): https://github.com/ibi-group/datatools-ui/pull/640/checks?check_run_id=1927768338

Yeah it's failing, but it's not even getting to the point where the tests start running. The test setup script has indicated that some environment variables don't seem to have been setup properly.

evansiroky (changes requested):
> @evansiroky, I'm curious if you have any insights on why things might be failing.

I fixed a few things related to the MS Teams webhook. There is still the matter of setting the AWS credentials to upload the e2e logs, but once that is done, the logs should get uploaded to AWS and we'll have a nice link to download those logs in the message that gets sent to the webhook. It's late out here, but I think getting the e2e logs uploaded to s3 is all that remains to be completed as far as this PR is concerned. Yes, the e2e tests aren't making it all the way through, but that's a separate concern and the main point is that the e2e tests actually run and any logs from the tests (almost) get uploaded so we can investigate if desired.

.github/workflows/node-ci.yml
```diff
@@ -0,0 +1,120 @@
...
+        # Set health checks to wait until postgres has started
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python
+        uses: actions/setup-python@v2
```
binh-dam-ibigroup, line 29: I just came across [this AWS announcement regarding Python versions](https://docs.aws.amazon.com/cli/latest/userguide/welcome-versions.html). It is not a blocking item, however if we proceed as written, we should also open an issue to install Python 3.6+.

binh-dam-ibigroup: Disregards, the v2 tag on the action is misleading...

landonreed:
@evansiroky, e2e is now fully working, see https://github.com/ibi-group/datatools-ui/runs/2096422003 for the successful run.

.github/workflows/node-ci.yml
```diff
@@ -0,0 +1,120 @@
...
+      - name: Cache pip
+        uses: actions/cache@v2
+        with:
+          # This path is specific to Ubuntu
+          path: ~/.cache/pip
+          # Look to see if there is a cache hit for the corresponding requirements file
+          key: ${{ runner.os }}-pip-${{ hashFiles('requirements.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-pip-
+            ${{ runner.os }}-
```
binh-dam-ibigroup, line 39: Have you tried without these two steps? According to the link in [your comment here](https://github.com/conveyal/transitive.js/pull/63#discussion_r593242095), Python, pip, and even node come preinstalled on Github's `ubuntu-latest`.

landonreed: Great question! I forget! (But will check.) As for removing the node install, I'm a bit on the fence about that one, since we don't necessarily want that version changing on us willy nilly.

landonreed: Looks like we don't actually need to install these! https://github.com/ibi-group/datatools-ui/pull/640/commits/bae498881934d6305eb4de2e315482441d89e92b

evansiroky (approved):
Perfect! I see the perfectly passing report on MS Teams and a nice codecov report segmented out by unit test vs e2e test.

PR #641, opened 2021-02-17
Otp 2 Refactor

landonreed:
This PR contains changes for #638. Primarily, it just breaks out the massive DeploymentViewer file into single component files. It also tweaks the styling/layout for custom file button toolbar.

I had considered possibly consolidating code that handles custom router/build config with the custom files, but there didn't seem to be a good way.

### DIFF
--- a/lib/manager/components/deployment/CustomFileEditor.js
+++ b/lib/manager/components/deployment/CustomFileEditor.js
@@ -1,0 +1,256 @@
+// @flow
+
+import React, {Component} from 'react'
+import {
+  Button,
+  ButtonToolbar,
+  Checkbox,
+  ControlLabel,
+  FormControl,
+  FormGroup,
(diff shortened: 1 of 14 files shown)


lib/manager/components/deployment/CustomFileEditor.js
```diff
@@ -0,0 +1,256 @@
...
+            ? <FormControl
+              onChange={this._onChangeFilename}
+              placeholder='Enter filename'
+              type='text'
+              value={customFile.filename}
+            />
+            : <span>
+              Filename:{' '}
+              <span style={{fontFamily: 'monospace'}}>
+                {/* FIXME: what should the behavior be here if missing */}
```
evansiroky, line 194: There should be a validation error if the filename is missing when raw contents are provided. I will fix this in my PR.

PR #644, opened 2021-03-02
Navigate to latest feed version after processing complete

### LINKED ISSUE #643: UI should display merged feed version after merging (MTC)

After merging a feed version with another, datatools does not display the resulting merged feed version.

**Exception:** If the URL when performing a merge does not end in `/version/<n>`, then datatools does show the resulting version after merge.

## Expected behavior

Datatools should show the merged feed version after a merge.

1. View a feed
2. Toggle between different versions.
3. Perform a merge
4. Datatools stays on the same on the feed version.

Configuration: MTC.

Recent dev branch.

binh-dam-ibigroup:
### Checklist

### Description

This PR fixes #643.

Under the hood, the PR changes job handling so that if viewing a particular feed, datatools navigates to a new feed version when that version's processing is complete. Hopefully that does not create undesired side effects.

### DIFF
--- a/lib/manager/actions/feeds.js
+++ b/lib/manager/actions/feeds.js
@@ -141,14 +141,6 @@ export function fetchFeedSource (feedSourceId: string) {
         dispatch(fetchRelatedTables(feedSource))
         return feedSource
       })
-      .then(() => {
-        // If viewing a particular feed, navigate to a new feed version as soon as it becomes available.
-        // (If user is not looking at this feed, don't navigate away from their current page.)
-        const newVersionPath = `/feed/${feedSourceId}`
-        if (browserHistory.getCurrentLocation().pathname.startsWith(`${newVersionPath}/version/`)) {
-          browserHistory.push(newVersionPath)
-        }
-      })
   }
 }
 
--- a/lib/manager/actions/feeds.js
+++ b/lib/manager/actions/feeds.js
@@ -141,6 +141,14 @@ export function fetchFeedSource (feedSourceId: string) {
         dispatch(fetchRelatedTables(feedSource))
         return feedSource
       })
+      .then(() => {
+        // If viewing a particular feed, navigate to a new feed version as soon as it becomes available.
+        // (If user is not looking at this feed, don't navigate away from their current page.)
+        const newVersionPath = `/feed/${feedSourceId}`
+        if (browserHistory.getCurrentLocation().pathname.startsWith(`${newVersionPath}/version/`)) {
+          browserHistory.push(newVersionPath)
+        }
+      })
   }
 }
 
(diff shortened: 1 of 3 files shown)


lib/manager/actions/status.js
```diff
@@ -259,6 +260,17 @@ export function handleFinishedJob (job: ServerJob) {
...
+      case 'PROCESS_FEED':
+        // If viewing a particular feed, navigate to a new feed version as soon as it becomes available.
+        // It is the same path (without version number) as the link shown for this job status.
+        // (If user is not looking at this feed, don't navigate away from their current page.)
+        if (job.feedSourceId) {
+          const newVersionPath = `/feed/${job.feedSourceId}`
+          if (browserHistory.getCurrentLocation().pathname.startsWith(newVersionPath)) {
+            browserHistory.push(newVersionPath)
+          }
+        }
```
landonreed, line 272: This generally looks good, a couple of comments:
- there is already logic in the `VALIDATE_FEED` case that fetches the feedSource. I think this check/browser push might be better if it were handled in a `then()` method following that fetch.
- there is a (tiny) bit of nuance missing in this check. If we're already on the feed source path, we don't really need to push to that path (the new version will load automatically). I think what we should be checking here is if someone is viewing a page that startsWith `/feed/${job.feedSourceId}/version`, then we push to `/feed/${job.feedSourceId}`.

landonreed (changes requested):
Looks good! Just a couple of small tweaks.

lib/manager/actions/status.js
binh-dam-ibigroup: Updated in https://github.com/ibi-group/datatools-ui/pull/644/commits/ae230d68aa4b3b1b97d137f9ba63f5a036996e8f. You're right about the check, I intended to do that!

lib/manager/actions/feeds.js
```diff
@@ -141,6 +141,14 @@ export function fetchFeedSource (feedSourceId: string) {
...
         return feedSource
       })
+      .then(() => {
+        // If viewing a particular feed, navigate to a new feed version as soon as it becomes available.
+        // (If user is not looking at this feed, don't navigate away from their current page.)
+        const newVersionPath = `/feed/${feedSourceId}`
+        if (browserHistory.getCurrentLocation().pathname.startsWith(`${newVersionPath}/version/`)) {
+          browserHistory.push(newVersionPath)
+        }
+      })
```
landonreed, line 151: Oops, sorry for the confusion, I meant that this should go in the `then()` when `dispatch(fetchFeedSource())` is called in the switch statement in `actions/status.js`.

binh-dam-ibigroup: Alright, updated in https://github.com/ibi-group/datatools-ui/pull/644/commits/e523505e7713ae582879443a8c57a0bafd70cf00.

landonreed (approved):

evansiroky (approved):

PR #649, opened 2021-03-22
Add normalize field transformation

binh-dam-ibigroup:
### Checklist

### Description

This PR adds UI features to normalize field names on processing a new GTFS zip file:
- Capitalize field names, with exceptions
- Substitute text/words with other text (and normalize space if replacing words).

Companion PR: https://github.com/ibi-group/datatools-server/pull/371.

Screenshot (updated):

image


evansiroky (changes requested):
Before submitting for review, can you please merge `dev` into this branch so that GitHub Actions are ran?

lib/manager/components/transform/NormalizeField.js
```diff
@@ -0,0 +1,223 @@
...
+  _onSaveTransformationSettings = () => {
+    this.props.onSave(this.state, this.props.index)
+  }
+
+  componentDidUpdate (prevProps: Props) {
+    const { transformation } = this.props
+    if (!isEqual(transformation, prevProps.transformation)) {
+      this.setState(getTransformationFields(transformation))
+    }
+  }
```
evansiroky, line 108: This method isn't recommended according to a [React blog post](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#recommendation-fully-uncontrolled-component-with-a-key). To reset some values, one recommended option is to pass in a `key` prop that will be different whenever the transformation value is different.

lib/manager/components/transform/ReplaceFileFromString.js
```diff
@@ -0,0 +1,96 @@
...
+ */
+export default class ReplaceFileFromString extends Component<Props, {csvData: ?string}> {
+  state = { csvData: null }
+
+  componentDidUpdate (prevProps: Props) {
+    const { transformation } = this.props
+    if (!isEqual(transformation, prevProps.transformation)) {
+      this.setState({ csvData: transformation.csvData })
+    }
+  }
```
evansiroky, line 41: This method isn't recommended according to a [React blog post](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#recommendation-fully-uncontrolled-component-with-a-key). To reset some values, one recommended option is to pass in a `key` prop that will be different whenever the transformation value is different.

lib/manager/components/transform/ReplaceFileFromString.js
```diff
@@ -0,0 +1,96 @@
...
+
+  render () {
```
evansiroky, line 43: I see that these methods are alphabetized, but I kind of like it when `render` is the last method. It's easier to find that way.

lib/manager/components/transform/FeedTransformation.js
```diff
@@ -23,124 +25,56 @@ type Props = {
...
-        // TODO: Further validate CSV data based on parsing fields, checking for
-        //   missing columns, etc.
-        break
-      default:
-        break
+
+    const {getValidationIssues} = transformationTypes[transformation['@type']]
+    if (typeof getValidationIssues === 'function') {
+      issues = issues.concat(getValidationIssues(transformation, feedSource))
     }
```
evansiroky, line 74: This sends over the `transformation` prop and not the values in the current state of the transformation. So if the user edits the NormalizeFieldTransformation for example, the validation issues won't go away until the transformation is persisted.

evansiroky (changes requested):
See comments, but only the comment about the validation issues is blocking.

lib/manager/components/transform/NormalizeField.js
binh-dam-ibigroup: Good suggestion, updated in https://github.com/ibi-group/datatools-ui/pull/649/commits/a7296121781a1413670f4e5ad6eb4bafabbee7bb.

lib/manager/components/transform/ReplaceFileFromString.js
binh-dam-ibigroup: Updated in https://github.com/ibi-group/datatools-ui/pull/649/commits/a7296121781a1413670f4e5ad6eb4bafabbee7bb.

lib/manager/components/transform/ReplaceFileFromString.js
binh-dam-ibigroup: Updated in https://github.com/ibi-group/datatools-ui/pull/649/commits/a7296121781a1413670f4e5ad6eb4bafabbee7bb.

lib/manager/components/transform/FeedTransformation.js
binh-dam-ibigroup: Thanks for pushing this. I moved validation code of the editing state inside each transform component in https://github.com/ibi-group/datatools-ui/pull/649/commits/ccc561232dfc744f574b02b6b35c44cdf0a73f9b.

binh-dam-ibigroup:
Assigning this back to myself to implement substitutions as a structure list.

binh-dam-ibigroup:
Commit https://github.com/ibi-group/datatools-ui/pull/649/commits/67a416c654f106bee6074525549ae7aa1d2e3884 replaces the funky `=>`/`=>*` syntax and now matches the structure requested in https://github.com/ibi-group/datatools-server/pull/371#discussion_r617092209.

lib/types/index.js
```diff
@@ -291,16 +291,38 @@ export type RetrievalMethod = $Values<typeof RETRIEVAL_METHODS>
...
   matchField?: string,
   matchValues?: Array<string>,
   sourceVersionId?: string,
   table: ?string
 }
 
+export type FeedTransformation =
+  & FeedTransformationBase
+  & ReplaceFileFromStringFields
+  & NormalizeFieldFields
```
evansiroky, line 324: I'd prefer if each transformation type extended the `FeedTransformationBase` and then this `FeedTransformation` type would be a [union type](https://flow.org/en/docs/types/unions/) of all the possible transformations.

lib/manager/components/transform/Transform.js
```diff
@@ -0,0 +1,63 @@
...
+  index: number,
+  onSave: (StateType, number) => void,
+  onValidationErrors: (Array<string>) => void,
+  transformation: FeedTransformationType
+}
+
+/**
+ * Generic Transform component that holds helper methods for derived components.
+ */
+export default class Transform<StateType> extends Component<TransformProps<StateType>, StateType> {
```
evansiroky, line 21: [Inheritance is not recommended for creating React components](https://reactjs.org/docs/composition-vs-inheritance.html). In here it seems like the `updateState` is the main way components communicate with this component which then will call `onValidationErrors`. Perhaps one way to go to avoid inheritance is to create a component that handles all of the differences between each transformation type. But by then this all might start to look like the beginning of a form state management component which kind of brings up the question of why aren't we using something like Formik. Also, maybe things would be easier if all these transformRules were a separate mongo collection instead of a list within a feed source record.

evansiroky (changes requested):
This addition of components using inheritance is giving me red flag vibes. Something doesn't seem right here and a decent refactor may be needed to simplify things.

lib/manager/components/transform/NormalizeField.js
```diff
@@ -0,0 +1,339 @@
...
+  index: number,
+  onChange: (Substitution, number) => void,
+  onRemove: number => void,
+  substitution: Substitution
+}
+
+/**
+ * Renders and lets user edit a substitution.
+ */
+class SubstitutionRow extends Component<SubstitutionRowProps> {
```
landonreed, line 275: Please place this component in a separate file. This file is too large.

lib/manager/components/transform/Transform.js
landonreed: Yea, this seems like a bit of a radical change to the structure of the components that might have benefitted from a technical discussion before taking the plunge.

lib/manager/components/transform/FeedTransformation.js
```diff
@@ -189,6 +114,7 @@ export default class FeedTransformation extends Component<Props, {csvData: ?stri
           <Select
             style={{margin: '10px 0px', width: '230px'}}
             clearable={false}
+            searchable={false}
```
landonreed, line 117: Why make this unsearchable?

lib/manager/components/transform/NormalizeField.js
```diff
@@ -0,0 +1,339 @@
...
+  const table = getGtfsTableSpec(tableName)
+  if (!table) return null
+
+  return (
+    <Select
+      clearable={false}
+      onChange={onChange}
+      options={table.fields.map(field => ({label: field.name, value: field.name}))}
+      placeholder='Choose the field to normalize'
+      searchable={false}
```
landonreed, line 258: I don't think this should be unsearchable.

lib/manager/components/transform/NormalizeField.js
```diff
@@ -0,0 +1,339 @@
...
+    throw new Error('GTFS spec could not be found!')
+  }
+  const tableFile = `${tableName}.txt`
+  return spec.find(t => t.name === tableFile)
+}
+
+/**
+ * Dropdown selector to select a field from a given GTFS table.
+ */
+const GtfsFieldSelector = (props: GtfsFieldSelectorProps) => {
```
landonreed, line 245: I could see this getting reused, so it would be nice to place it in its own file.

lib/manager/components/transform/ReplaceFileFromString.js
```diff
@@ -0,0 +1,82 @@
...
+import Transform from './Transform'
+
+import type { ReplaceFileFromStringFields } from '../../../types'
+import type { TransformProps } from './Transform'
+
+/**
+ * Component that renders fields for ReplaceFileFromStringTransformation.
+ */
+export default class ReplaceFileFromString extends Transform<ReplaceFileFromStringFields> {
+  constructor (props: TransformProps<ReplaceFileFromStringFields>) {
```
landonreed, line 15: I agree with @evansiroky about avoiding inheritance... it just takes us away from patterns that are used everywhere else in our apps and makes it tough for the reader to switch contexts.

landonreed (changes requested):
See comments about inheritance, making files smaller, and issue saving field selection on QA server (mentioned on Teams).

lib/manager/components/transform/NormalizeField.js
binh-dam-ibigroup: Moved in https://github.com/ibi-group/datatools-ui/pull/649/commits/19fb3d2f7d2a415b941f3b47e667877c0958fe69.

lib/manager/components/transform/ReplaceFileFromString.js
binh-dam-ibigroup: Addressed in https://github.com/ibi-group/datatools-ui/pull/649/commits/54f1aac3573cc5af4575bd839afb116708e626e8

lib/manager/components/transform/Transform.js
binh-dam-ibigroup: I changed that in https://github.com/ibi-group/datatools-ui/pull/649/commits/54f1aac3573cc5af4575bd839afb116708e626e8, although it creates some code duplication for managing the state.

lib/types/index.js
binh-dam-ibigroup: Updated in https://github.com/ibi-group/datatools-ui/pull/649/commits/54f1aac3573cc5af4575bd839afb116708e626e8. There is one prop `typeName` I had to add that duplicates `@type`, without which flow rejects the case in `getTransformationName` that involves `csvData`. (Flow does not seem to accept attribute `@type` for type branching)

lib/manager/components/transform/NormalizeField.js
binh-dam-ibigroup: This was done in https://github.com/ibi-group/datatools-ui/pull/649/commits/7778d4bb3123d39936be338917e1331f1f93a89b.

lib/manager/components/transform/NormalizeField.js
binh-dam-ibigroup: This is done as part of in https://github.com/ibi-group/datatools-ui/pull/649/commits/7778d4bb3123d39936be338917e1331f1f93a89b.

lib/manager/components/transform/FeedTransformation.js
binh-dam-ibigroup: Reverted in https://github.com/ibi-group/datatools-ui/pull/649/commits/9a4a01d4b5a48e77d70e5e34c9f20226a04a2e8e.

evansiroky (approved):
Great stuff, @binh-dam-ibigroup. I think the slightly extra amount of duplicate code vs adding inheritance will keep things simpler and easier to maintain.

lib/manager/components/transform/ReplaceFileFromVersion.js
```diff
@@ -35,6 +35,21 @@ export default class ReplaceFileFromVersion extends Transform<{}> {
...
+  _updateErrors = () => {
+    const { onValidationErrors } = this.props
+    if (typeof onValidationErrors === 'function') {
+      onValidationErrors(this.getValidationErrors())
+    }
+  }
+
+  componentDidMount () {
+    this._updateErrors()
+  }
```
landonreed, line 51: Generally, I think the pattern we follow is to place lifecycle methods like this at the beginning of the component. Same comment goes for the other files in this PR.

landonreed (approved):
Looks great! Approved once #681 is merged.

binh-dam-ibigroup:
Ready for review again. Comment https://github.com/ibi-group/datatools-ui/pull/681#issuecomment-840832999 is addressed by (1) https://github.com/ibi-group/datatools-ui/pull/649/commits/c87759946e6f4063138a115f870aa16d28c5e94a and (2) https://github.com/ibi-group/datatools-ui/pull/649/commits/9f10808f1536bc3c6eac246b38ff77106f211e42 / https://github.com/ibi-group/datatools-server/pull/371/commits/6fb9a0ffb65deeaeb979311be04d70f702813abc.

lib/manager/components/transform/NormalizeField.js
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  _onRemoveSubstitution = (index: number) => {
+    const newSubstitutions = [].concat(this.props.transformation.substitutions)
+    newSubstitutions.splice(index, 1)
+    this._updateTransformation({substitutions: newSubstitutions})
+  }
+
+  _onNewSubstitution = () => {
+    this._updateTransformation({substitutions: [...[].concat(this.props.transformation.substitutions), {
```
evansiroky, line 59: Given a new transformation, when this is executed, then the `this.props.transformation.substitutions` is null which causes the array to become `[null]` which leads to a downstream error anywhere that properties on this supposed expected substitution object are checked. This results in the Add button not adding any substitutions and just error'ing out.

evansiroky (changes requested):
See comment.

lib/manager/components/transform/NormalizeField.js
binh-dam-ibigroup: Thanks for finding this bug! Fixed in. https://github.com/ibi-group/datatools-ui/pull/649/commits/a595d1a6b4f0f173dc330fb9987f0489e49d20ee.

evansiroky (changes requested):
Some oddities I have observed with adding substitutions:

- The placeholders seem to be the actual values upon adding a new rows
- Sometimes if you click "Add" (substitution) with just one row immediately after editing the replacement without blurring the replacement field, no new row is created
- If you click "Add" with more than one row after about 1 second of editing a replacement (without blurring), a new row will be added but the change made to the replacement will be lost

I believe all of these are a result of dealing with a list that is stored within a mongo object. This result in concurrency issues because whenever this list is attempted to be updated, updates in parallel cannot occur otherwise a race condition will exist where one version of the list will win and the updates from the other will be lost.

The solution to this problem could be to either lock editing to only one specific list item at a time or to make the list a new collection within the database that has its own IDs that would enable asynchronous editing.

However, thinking further, this really doesn't do anything about the use case of two users simultaneously editing the same list. But at the very least, I can't approve until at the very least one of the above two solutions is implemented to at least solve the asynchronous editing problem for a single user editing the transformation.

binh-dam-ibigroup:
> * The placeholders seem to be the actual values upon adding a new rows
Correct, they are templates as written. Should they be left blank?

> * Sometimes if you click "Add" (substitution) with just one row immediately after editing the replacement without blurring the replacement field, no new row is created
> * If you click "Add" with more than one row after about 1 second of editing a replacement (without blurring), a new row will be added but the change made to the replacement will be lost
> 
> I believe all of these are a result of dealing with a list that is stored within a mongo object. This result in concurrency issues because whenever this list is attempted to be updated, updates in parallel cannot occur otherwise a race condition will exist where one version of the list will win and the updates from the other will be lost.
As written, there is indeed a race condition, but the behavior you observed comes from the full remounting/rerendering of `NormalizeField` after an edit is made and the persisted values are refetched from the backend. I changed the UI component in https://github.com/ibi-group/datatools-ui/pull/649/commits/6e74f4859f36769af29febb7c4cd3cea96f5475a to queue the add/delete operations, so they execute after the edits take effect.

evansiroky (changes requested):
This definitely will not be able to properly apply updates if there is any noticeable amount of latency with the network connection and a user who edits things really rapidly. Try setting the [network throttling to "fast 3G" in Chrome Dev tools](https://developer.chrome.com/docs/devtools/network/reference/#throttling) and then edit a few different fields as if you were a power-editor. I really recommend implementing a change that would lock an entire record and have a edit/save button which at the very least will slow down the user editing to make sure we aren't sending over data to be saved that isn't built on top of the latest version of that record in the database. I kind of did this when implementing the [custom file editor](https://github.com/ibi-group/datatools-ui/blob/dev/lib/manager/components/deployment/CustomFileEditor.js) in the deployment.

binh-dam-ibigroup:
> This definitely will not be able to properly apply updates if there is any noticeable amount of latency with the network connection and a user who edits things really rapidly. Try setting the [network throttling to "fast 3G" in Chrome Dev tools](https://developer.chrome.com/docs/devtools/network/reference/#throttling) and then edit a few different fields as if you were a power-editor. I really recommend implementing a change that would lock an entire record and have a edit/save button which at the very least will slow down the user editing to make sure we aren't sending over data to be saved that isn't built on top of the latest version of that record in the database. I kind of did this when implementing the [custom file editor](https://github.com/ibi-group/datatools-ui/blob/dev/lib/manager/components/deployment/CustomFileEditor.js) in the deployment.

@evansiroky Save/revert functionality was available until https://github.com/ibi-group/datatools-ui/pull/649/commits/c87759946e6f4063138a115f870aa16d28c5e94a by suggestion of @landonreed. Given the added unnecessary network traffic inline edits create, I am in favor of going back to a Save/revert functionality. But then, for consistency, table selector should also be converted to work that way, so there is a potentially large refactor. Thoughts?

binh-dam-ibigroup:
I made changes as discussed during our call, see commits https://github.com/ibi-group/datatools-ui/pull/649/commits/1d483020b1a4314b3c85f1a755437c95fc476f50 and following.

landonreed (changes requested):
@binh-dam-ibigroup this looks really good! A couple of tweaks on the UX:
- it's not obvious that you can click a row to begin editing. It might help users know this is ok by adding `cursor: pointer` to the style when "click to edit" is allowed.
- We should probably just prohibit users from saving a row if it is invalid. Right now, I can clear the search text and click "save", but only then am I shown the validation issue.
- when I click "Add", I kind of expect the added row to be in the "editable" state, rather than having to also click "Edit".

I haven't looked at the code changes yet, but wanted to get your eyes on these first.

binh-dam-ibigroup:
> * We should probably just prohibit users from saving a row if it is invalid. Right now, I can clear the search text and click "save", but only then am I shown the validation issue.
Good idea.

> * when I click "Add", I kind of expect the added row to be in the "editable" state, rather than having to also click "Edit".
> 
I initially felt the same about it. One side effect: that action will prevent the user to add two new rows, they would have to save or dismiss a newly created row first before they can add a new one. Do we want that?

landonreed:
> One side effect: that action will prevent the user to add two new rows, they would have to save or dismiss a newly created row first before they can add a new one. Do we want that?

Yes, I think that side effect is fine. Also, I'm not really sure it's valuable to pre-populate the value of the search/replace with fields.

binh-dam-ibigroup:
Ready for review again, see the latest tweaks from https://github.com/ibi-group/datatools-ui/pull/649/commits/5839fc9a13a9587531d57306d585e207086374c9.

evansiroky (approved):
Approving so we can move along, but I still think there are 2 design problems that we should seriously consider avoiding in the future.

1. The decision to make an a List of individually-editable entities within a record of another parent object. We should seriously consider avoiding creating a list field within a record unless all entities must be edited at once. We should also consider refactoring the existing cases where this occurs.
2. The implementation of a custom form made with code. We should consider sticking to a form library like Formik that will reduce a lot of this boilerplate. Furthermore, using a single form library will standardize the look and feel of other places that have data entry elsewhere in the application. In this PR we went back and forth a lot before finally coming to a decision on how to create the inputs. If we had used a standard single form library, there would be an expectation of how to implement form inputs and therefore the coding and review time should be shorter.

landonreed (changes requested):
- I think things would be easier/cleaner if we defaulted the substitutions list to an empty list (rather than null)
- There's a minor bug in the ReplaceFileFromVersion transformation causing a validation issue to show when it is not relevant (doesn't appear to be a problem with ReplaceFileFromString):
image

binh-dam-ibigroup:
> * There's a minor bug in the ReplaceFileFromVersion transformation causing a validation issue to show when it is not relevant (doesn't appear to be a problem with ReplaceFileFromString):
>   image

Good find, this is fixed in https://github.com/ibi-group/datatools-ui/pull/649/commits/7b00cee840ea112c6961167dc26b613ec040ddc8.

binh-dam-ibigroup:
> * I think things would be easier/cleaner if we defaulted the substitutions list to an empty list (rather than null)

That comes from the backend hence the null check. I refactored that null check in https://github.com/ibi-group/datatools-ui/pull/649/commits/fadcc23808f74c0321112eb6997242c900ed665a.

binh-dam-ibigroup:
The latest changes from https://github.com/ibi-group/datatools-ui/pull/649/commits/7897565431e26ddcb393990c0b00951fac42a6ac require https://github.com/ibi-group/datatools-server/pull/386.

landonreed (approved):

PR #656, opened 2021-04-08
Added missing GTFS Extended Route Types

### LINKED ISSUE #645: Support for Extended GTFS Route Types
_**NOTE:** This issue system is intended for reporting bugs and tracking progress in software development. Although this software is licensed with an open-source license, any issue opened here may not be responded to in a timely manner. [Conveyal](https://www.conveyal.com) is unable to provide technical support for custom deployments of this software unless your company has a support contract with us. Please remove this note when creating the issue._

No support for creating and editing routes with Extended GTFS Route Types

## Expected behavior

Support for editing Extended GTFS Route Types

I would like to create 1500 Water Taxi Service and 1000 Water Transport Service route types. 

none

UI Version: | 872d61
Server version: | 11e9a8

Is the support for Extended GTFS Route Types on the roadmap? Or is as simple as editing the .yml config files and creating a pull request for this?

robertgregg3:
- Added the route types to the gtfs.yml file 
- ~and then added logic to display the route_type field as an optgroup~
- Create a new selector for GTFS route types, with extended route types organized as a tree hierarchy (bus, metro, etc) that is not expanded by default to reduce screen clutter while searching for extended route types.

Fix #645


lib/editor/components/EditorInput.js
```diff
@@ -300,6 +300,8 @@ export default class EditorInput extends Component<Props> {
           </span>
         )
       case 'DROPDOWN':
+        const standard = 'Standard'
+        const extended = 'Extended'
         return (
```
landonreed, line 305: To fix the flow type issues and simplify things, I would add something like the following:
```javascript
if (!field.options) {
  console.warn(`Field `field.name` does not have options configured, but is marked as inputType=DROPDOWN`)
  return null
}
```

gtfs.yml
```diff
@@ -290,6 +290,156 @@
       text: Gondola
     - value: 7
       text: Funicular
```
landonreed, line 292: Could you also add options 11 and 12 (they were recently added to the standard set)? See the spec for `routes.txt`: https://developers.google.com/transit/gtfs/reference#routestxt for more info

lib/editor/components/EditorInput.js
```diff
@@ -314,7 +316,18 @@ export default class EditorInput extends Component<Props> {
...
+                <optgroup label={standard}>
+                  {field.options.slice(0, 7).map(o => (
+                    <option value={o.value} key={o.value}>{o.text} ({o.value})</option>))}
+                </optgroup>}
+              {field.name === 'route_type' &&
+                <optgroup label={extended}>
+                  {field.options.slice(8).map(o => (
+                    <option value={o.value} key={o.value}>{o.text} ({o.value})</option>))}
+                </optgroup>}
+              )}
```
landonreed, line 329: I would recommend placing this stuff (including lines 303 and 304) into a `_renderRouteTypeOptions` method to encapsulate it a bit more. Once that's done, I would recommend you use a ternary here instead of making the check for field name twice:
```javascript
field.name === 'route_type'
  ? this._renderRouteTypeOptions(options)
  : field.options.map(o => (<option value={o.value} key={o.value}>{o.text || o.value}</option>))
```

landonreed (changes requested):
A few changes, but this is looking good!

lib/editor/components/EditorInput.js
robertgregg3: Where do you see the flow issues?  When I load the project (before typing any code, I get these kind of squigglies from the beginning of the render method:
image

robertgregg3: When I run `flow status` I get an error:
image

gtfs.yml
robertgregg3: Done!

gtfs.yml
```diff
@@ -290,6 +290,156 @@
...
+    - value: 100
+      text: Railway Service
```
binh-dam-ibigroup, line 294: Why is this value duplicated?

robertgregg3: Good spot. And there was a word missing. Going to check the rest now.

lib/editor/components/EditorInput.js
landonreed: The flow errors are showing up in the GitHub builds: https://github.com/ibi-group/datatools-ui/pull/656/checks?check_run_id=2298915284

Locally, you should run `yarn flow`.

lib/editor/components/EditorInput.js
```diff
@@ -314,7 +316,18 @@ export default class EditorInput extends Component<Props> {
                 ? <option disabled value=''>{field.required ? '-- select an option --' : '(optional)' }</option>
                 : null
               }
-              {field.options && field.options.map(o => (<option value={o.value} key={o.value}>{o.text || o.value}</option>))}
+              {field.name === 'route_type' &&
+                <optgroup label={standard}>
+                  {field.options.slice(0, 7).map(o => (
```
binh-dam-ibigroup, line 321: Because `o` shows up so many times, I would write `.map(({ text, value }) => { ...`, and maybe also extract the `<optgroup>` rendering to its own method because of repetition.

robertgregg3:
### I have added:
- a group option in the gtfs file
- added a _renderRouteTypeOptions method to dynamically render the opt groups depending on the group value (Standard or Extended)

landonreed (commented):
@robertgregg3, what prompted the addition of the `group` field in `gtfs.yml`? I'm not sure it's needed, given that we effectively know that types 0-12 are standard and those after are extended. I liked the way you had handled it previously.

Also, `Trolleybus (11)` and `Monorail (12)` should probably be in standard and moved up in the list. These are currently shown as extended (800 and 405), but I think that's probably just because the extended list has not been updated. We could keep them both, but that might be confusing.

lib/editor/components/EditorInput.js
```diff
@@ -118,6 +118,33 @@ export default class EditorInput extends Component<Props> {
...
+    const extendedTypes = []
+    const routeTypes = ['Standard', 'Extended']
+    const fieldOptions = field.options
+    for (const option in fieldOptions) {
+      if (fieldOptions[option].group === 'Standard') {
+        standardTypes.push(fieldOptions[option])
+      } else {
+        extendedTypes.push(fieldOptions[option])
+      }
+    }
```
binh-dam-ibigroup, line 132: This could be simplified:
```js
const standardTypes = Object.values(fieldOptions).filter(fieldOption => fieldOption.group === 'Standard')
// same for extendedTypes
```

binh-dam-ibigroup: Even better, in the `return` statement, try this:
```js
return routeTypes
.map(type => Object.values(fieldOptions).filter(/*see above*/))
.map(routeTypes => {
  <optgroup>
    {routeTypes.map(routeType => /*what you already have for each std/extended types*/)}
  </optgroup>
})

**Review by binh-dam-ibigroup — CHANGES_REQUESTED** [review]
Almost there, see how the simplification suggestion would work with @landonreed's question regarding the groups of route types.

**Comment by landonreed** [comment]
Commentary is a little muddy here, so I'm going to add a PR and share my recs there.

**Comment by mvanlaar** [comment]
> @robertgregg3, what prompted the addition of the `group` field in `gtfs.yml`? I'm not sure it's needed, given that we effectively know that types 0-12 are standard and those after are extended. I liked the way you had handled it previously.
> 
> Also, `Trolleybus (11)` and `Monorail (12)` should probably be in standard and moved up in the list. These are currently shown as extended (800 and 405), but I think that's probably just because the extended list has not been updated. We could keep them both, but that might be confusing.

My opinion on this would be there will no code changes needed in the future if somethings changes to the spec. Only a edit to the gtfs.yml file. Also some companies can changes which items are for them standard and what are extended without changing code but only the gtfs.yml file.

**Comment by landonreed** [comment]
@robertgregg3, take a look at my changes in https://github.com/ibi-group/datatools-ui/pull/659. Feel free to review and merge here if they look OK to you. And let me know if you have any questions!

**Comment by robertgregg3** [comment]
> @robertgregg3, what prompted the addition of the `group` field in `gtfs.yml`? I'm not sure it's needed, given that we effectively know that types 0-12 are standard and those after are extended. I liked the way you had handled it previously.
> 
@landonreed I had an interesting time trying to create an drop down with multiple optgroups (in React).  The form control in the DROPDOWN case appears to be looking for one nested div of options. When I tried to render 2 opt groups within one drop down I was getting the JSX error about it needing a surrounding div tag or React.Fragment. 

When researching I cam across others with similar issues in React. I found a solution that used grouping via the options data so that is why I added the "group" to the gtfs.yml route_type options.  

I did kind of prefer the other option and there was most likely a simple face palm solution.

File: lib/editor/components/EditorInput.js (hunk 6 above)
Comment (robertgregg3) [reply lib/editor/components/EditorInput.js:132]: Nice, I like it! Thanks Binh

**Comment by robertgregg3** [comment]
@landonreed Thanks for sharing your solution. Looks great

**Comment by landonreed** [comment]
Re: @mvanlaar's point about avoiding a code change, yes, there may be some value in that. But these `route_type` values and their classification as Standard vs. Extended are intended to accurately represent the GTFS specification, rather than a third party's interpretation of what is "standard." I think we should stick with the slice method for now. Since both of these files (`gtfs.yml` and `EditorInput.js`) are committed to the repo (and there's no way to substitute alternative files for them), I don't see a major difference in the "no-code-change" (yml) vs. the js approach, so I'm inclined to stick with the current approach.

**Review by landonreed — APPROVED** [review]

**Comment by robertgregg3** [comment]
@binh-dam-ibigroup the changes are from Landons Merged commit

File: lib/editor/components/EditorInput.js [hunk 7]
```diff
@@ -1,6 +1,6 @@
 // @flow
 
-import React, {Component} from 'react'
+import * as React from 'react'
```
binh-dam-ibigroup, line 3: Can you use the previous import nomenclature? (unless others want to override me).

lib/editor/components/EditorInput.js
```diff
@@ -118,31 +118,26 @@ export default class EditorInput extends Component<Props> {
...
-    const standardTypes = []
-    const extendedTypes = []
-    const routeTypes = ['Standard', 'Extended']
-    const fieldOptions = field.options
-    for (const option in fieldOptions) {
-      if (fieldOptions[option].group === 'Standard') {
-        standardTypes.push(fieldOptions[option])
-      } else {
-        extendedTypes.push(fieldOptions[option])
+  _renderRouteTypeOptions = (field: GtfsSpecField): React.Node => {
```
binh-dam-ibigroup, line 121: I think the more accurate output type would be `Array<React.Node>` (non-blocking)

lib/editor/components/EditorInput.js
```diff
@@ -118,6 +118,28 @@ export default class EditorInput extends Component<Props> {
...
+    const options = field.options || []
+    const toOption = ({value, text}) => (
+      <option key={value}>{text} ({value})</option>
+    )
+    // The first 10 options (types 0 - 12) are the standard types found in the
+    // GTFS specification. All others represent extended route types.
+    const routeTypes = [
+      {
+        label: 'Standard',
+        options: options.slice(0, 9).map(toOption)
```
binh-dam-ibigroup, line 131: Using a filter on the range for `value` may be easier to read. As written, it looks like the array indexes and the comment above contradict each other.
```suggestion
        options: options.filter(opt => opt.value < 100).map(toOption)
```

lib/editor/components/EditorInput.js
```diff
@@ -118,6 +118,28 @@ export default class EditorInput extends Component<Props> {
...
+      },
+      {
+        label: 'Extended',
+        options: options.slice(10).map(toOption)
```
binh-dam-ibigroup, line 135:
```suggestion
        options: options.filter(opt => opt.value >= 100).map(toOption)
```

binh-dam-ibigroup (approved):
Conditional approval on using a filter for separating the standard vs. extended types for improved readability.

lib/editor/components/EditorInput.js
landonreed: No, because referencing React.Node requires this import approach.

lib/editor/components/EditorInput.js
landonreed:
```suggestion
  _renderRouteTypeOptions = (field: GtfsSpecField): Array<React.Node> => {
```

binh-dam-ibigroup:
@robertgregg3 Please fix the failing build...

binh-dam-ibigroup:
This PR seemed to be blocked pending #659, but those changes have been merged since. Removing the "Blocked" label.

binh-dam-ibigroup:
We can probably start reviewing this PR at this point. There may be some more refactorings possible that would touch parts that are not relevant to this PR, but let me know if you want to make them.

miles-grant-ibigroup (approved):
This actually looks quite good! The dropdown feels a bit janky but given that we're using a library this is perhaps to be expected. Other than a few style changes I think this might actually be good to go!

Thanks for updating the e2e tests as well.

lib/style.css
```diff
@@ -879,3 +879,111 @@ h4.line:after {
...
+  width: 100%;
+}
+
+.route-type-select .infinite-scroll-component {
+  overflow-x: hidden!important;
+}
+
+/* Remove border and background of tags
+ * (a "tag" is a selected item displayed inside the input portion of the selector). */
+.route-type-select .dropdown-trigger .tag {
```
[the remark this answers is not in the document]
binh-dam-ibigroup: Which bold are you referring to?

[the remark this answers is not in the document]
binh-dam-ibigroup: Oh I see, in FF.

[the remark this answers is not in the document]
binh-dam-ibigroup: Thanks, addressed in https://github.com/ibi-group/datatools-ui/pull/656/commits/bc14fedc2b8390f387d73beb60777a528e0bf857.

philip-cline (approved):
Looks good to me!

Issue #650, opened 2021-03-22
Change project selector UI to React Select


landonreed:
The current project selector on the home page is somewhat cumbersome to use when a user has access to many projects. Also, the "Create Project" option is at the very bottom of the dropdown, which makes for quite a chore to create a new project.

### Approach
- [ ] Replace the bootstrap DropdownButton for the selector with a React Select dropdown so that users can begin typing the project name and select an item more quickly.
- [ ] Remove entirely the "Manage Projects" option (it links to an unrelated page)
- [ ] Remove the "Create Project" option and instead replace with a New Project button to the right of the selector

### Current Selector UI
![image](https://user-images.githubusercontent.com/2370911/112031391-a8f0f280-8b11-11eb-9f32-3930c7e770fc.png)

### Mockup of new Approach
TODO

binh-dam-ibigroup:
I think the dropdown + button combo should be replaced with a project table that could be shown in the main area of the screen as opposed to the side.

robertgregg3:
@binh-dam-ibigroup  That would be good. What would normally go in the "Recent Activity section"?  Perhaps that can be moved to the sidebar depending on the importance.

PR #662, opened 2021-04-22
improvement(HomeProjectDropdown.js): Changed Bootstrap Dropdown to React-Select

robertgregg3:
#650

Swapped out the current dropdown box on the users homepage for a React-Select component.  The brief asked for 3 things:

1- To replace the drop down with React select
2 - To remove the "manage projects
3 - To remove the create project option and replace with a New project button

The default styling of the select box I thought looked nice full width so placed it below.  I felt the text button blended well without the need for another button so it did't look too "buttony" :-)

### DIFF
--- a/lib/manager/components/HomeProjectDropdown.js
+++ b/lib/manager/components/HomeProjectDropdown.js
@@ -56,7 +56,7 @@ export default class HomeProjectDropdown extends Component<Props, State> {
         />
         <LinkContainer to='/project/new' style={{listStyle: 'none'}}>
           <MenuItem>
-            <Icon type='plus' /> Create project
+            <Icon type='plus' /> New project
           </MenuItem>
         </LinkContainer>
       </div>
--- a/lib/manager/components/HomeProjectDropdown.js
+++ b/lib/manager/components/HomeProjectDropdown.js
@@ -2,9 +2,9 @@
 
 import Icon from '@conveyal/woonerf/components/icon'
 import React, {Component} from 'react'
-import { Button, DropdownButton, MenuItem } from 'react-bootstrap'
+import { Button, MenuItem } from 'react-bootstrap'
 import { LinkContainer } from 'react-router-bootstrap'
-import {getAbbreviatedProjectName} from '../../common/util/util'
+import Select from 'react-select'
 
 import type {Project} from '../../types'
 import type {ManagerUserState} from '../../types/reducers'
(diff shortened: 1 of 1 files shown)


lib/manager/components/HomeProjectDropdown.js
```diff
@@ -15,88 +15,50 @@ type Props = {
...
+  state = {
+    selectedOption: ''
+  };
+
+  handleChange = (selectedOption: any) => {
+    this.setState(
+      {selectedOption},
+      () => console.log(`Option selected:`, this.state.selectedOption)
+    )
+  };
```
landonreed, line 32: @robertgregg3, sorry if there was confusion about the "routing to the project page" discussion. The behavior of `onChange` handler should remain the same as the previous selector. I.e., it should route to `{`/home/${project.id}`}`, and should not use the component state.

lib/manager/components/HomeProjectDropdown.js
```diff
@@ -15,88 +15,50 @@ type Props = {
...
+          value={selectedOption}
+          style={{marginBottom: '5px', zIndex: '0'}}
+          onChange={this.handleChange}
+          options={visibleProjects.map(p => ({value: p.id, label: p.name}))} placeholder={`Select Project...`}
+        />
+        <LinkContainer to='/project/new' style={{listStyle: 'none'}}>
+          <MenuItem>
+            <Icon type='plus' /> New project
+          </MenuItem>
+        </LinkContainer>
```
landonreed, line 61: This should remain only accessible to admins (see the check that was removed for `isAdmin`).

lib/manager/components/HomeProjectDropdown.js
```diff
@@ -2,9 +2,9 @@
 
 import Icon from '@conveyal/woonerf/components/icon'
 import React, {Component} from 'react'
-import { Button, DropdownButton, MenuItem } from 'react-bootstrap'
+import { Button, MenuItem } from 'react-bootstrap'
 import { LinkContainer } from 'react-router-bootstrap'
-import {getAbbreviatedProjectName} from '../../common/util/util'
```
landonreed, line 7: Without this method things end up like this: 
image

lib/manager/components/HomeProjectDropdown.js
```diff
@@ -15,88 +15,50 @@ type Props = {
...
     if (!profile) return null
-    const abbreviatedProjectName = getAbbreviatedProjectName(activeProject)
     return (
       <div style={{marginBottom: '20px'}}>
         {activeProject
-          ? <LinkContainer className='pull-right' to={`/project/${activeProject.id}`}>
-            <Button bsStyle='primary'>View {abbreviatedProjectName}</Button>
+          ? <LinkContainer className='pull-right' style={{position: 'relative', listStyle: 'none', zIndex: '1', marginBottom: '5px'}} to={`/project/${selectedOption && selectedOption.value}`}>
+            <Button bsStyle='primary'>View {selectedOption && selectedOption.label}</Button>
           </LinkContainer>
```
landonreed, line 48: These lines are a bit too long, please reformat to be on multiple lines.

landonreed (changes requested):
@robertgregg3, just a few tweaks on this first pass for ya.

robertgregg3:
@landonreed thanks for the feedback. I will get to it!

robertgregg3:
Just noticed the 'clear' feature is not working

landonreed:
@robertgregg3 , just a heads up that builds are also failing.

robertgregg3:
@landonreed I'm not getting typing linting sqwigglies on OTP or ITMS. So will take a look at that tomorrow. Build fails have been all type related

lib/manager/components/HomeProjectDropdown.js
```diff
@@ -15,88 +17,55 @@ type Props = {
...
+}
+
+export default class HomeProjectDropdown extends Component<Props, State> {
+  state = {
+    selectedOption: ''
+  }
+
+  handleChange = (selectedOption: any, activeProject: any) => {
+    this.setState({selectedOption})
+    activeProject === null
```
binh-dam-ibigroup, line 31: I think there is some interference between variables `selectedOption` and `activeProject` in the animated steps below, where when the user clicks an entry after entering a plain `/home/` path, the selection is lost. Any reason to use two variables instead of one?

What I think should happen is, when the user selects an entry from the dropdown, the URL should be updated immediately, as in the current `dev` branch.

Screen Recording 2021-04-27 at 3 52 26 PM

lib/manager/components/HomeProjectDropdown.js
```diff
@@ -15,88 +17,55 @@ type Props = {
...
       activeProject,
       user,
       visibleProjects
     } = this.props
-    const isAdmin = user.permissions && user.permissions.isApplicationAdmin()
     const {profile} = user
-    if (!profile) return null
     const abbreviatedProjectName = getAbbreviatedProjectName(activeProject)
+    const isAdmin = user.permissions && user.permissions.isApplicationAdmin()
+    if (!profile) return null
```
binh-dam-ibigroup, line 45: Was there a need to move this line?

binh-dam-ibigroup (changes requested):
The layout looks good. Please advise regarding https://github.com/ibi-group/datatools-ui/pull/662/files#r621560348...

landonreed:
@robertgregg3, take a look at https://github.com/ibi-group/datatools-ui/pull/664. I've provided some comments in that PR, just to give you an idea of how I might have tweaked a couple of things and some of the underlying thinking. Feel free to merge if everything looks ok. And let me know if you have any questions at all!

robertgregg3:
@landonreed I used your solution and fixed one bug on top. Thanks for the help.

binh-dam-ibigroup (approved):

landonreed (approved):
Just finetuned the layout a bit and added custom renderers for React Select so that it matches the previous selector a bit. Also, I really liked your link button @robertgregg3, so I added that back in. I think this is good to go!

image

PR #664, opened 2021-04-27
refactor(HomeProjectDropdown): fix a couple minor things

landonreed:
A couple of tweaks for ya @robertgregg3. Figured this might be more helpful than overwhelming you with too many comments!

### DIFF
--- a/lib/manager/components/HomeProjectDropdown.js
+++ b/lib/manager/components/HomeProjectDropdown.js
@@ -18,8 +18,6 @@ type Props = {
   visibleProjects: Array<Project>
 }
 
-const toOption = (project: Project) => ({value: project.id, label: project.name})
-
 export default class HomeProjectDropdown extends Component<Props> {
   handleChange = (option: ReactSelectOption) => {
     browserHistory.push(`/home/${option ? option.value : ''}`)
@@ -35,7 +33,9 @@ export default class HomeProjectDropdown extends Component<Props> {
     const {profile} = user
     if (!profile) return null
     const abbreviatedProjectName = getAbbreviatedProjectName(activeProject)
-    const options = visibleProjects.map(toOption)
+    const options = visibleProjects.map(
+      project => ({value: project.id, label: project.name})
+    )
     return (
       <div style={{marginBottom: '20px'}}>
         {
--- a/lib/manager/components/HomeProjectDropdown.js
+++ b/lib/manager/components/HomeProjectDropdown.js
@@ -3,12 +3,13 @@
 import Icon from '@conveyal/woonerf/components/icon'
 import React, {Component} from 'react'
 import { Button } from 'react-bootstrap'
-import { LinkContainer } from 'react-router-bootstrap'
 import {browserHistory} from 'react-router'
+import { LinkContainer } from 'react-router-bootstrap'
 import Select from 'react-select'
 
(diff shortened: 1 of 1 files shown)


lib/manager/components/HomeProjectDropdown.js
```diff
@@ -3,12 +3,13 @@
...
 import { Button } from 'react-bootstrap'
-import { LinkContainer } from 'react-router-bootstrap'
 import {browserHistory} from 'react-router'
+import { LinkContainer } from 'react-router-bootstrap'
 import Select from 'react-select'
 
 import {getAbbreviatedProjectName} from '../../common/util/util'
-import type {Project} from '../../types'
+
+import type {Project, ReactSelectOption} from '../../types'
```
landonreed, line 12: We like to keep type imports separated from local project imports.

lib/manager/components/HomeProjectDropdown.js
```diff
@@ -17,20 +18,11 @@ type Props = {
   visibleProjects: Array<Project>
 }
 
-type State = {
-  selectedOption: any
-}
```
landonreed, line 22: Generally we try not to introduce local component state unless it's really warranted. Here, we're already tracking the selectedOption via the `activeProject`, which comes from the component's props (and ultimately from the URL that we update in handleChange).

lib/manager/components/HomeProjectDropdown.js
```diff
@@ -39,10 +29,13 @@ export default class HomeProjectDropdown extends Component<Props, State> {
...
     } = this.props
-    const {profile} = user
-    const abbreviatedProjectName = getAbbreviatedProjectName(activeProject)
     const isAdmin = user.permissions && user.permissions.isApplicationAdmin()
+    const {profile} = user
     if (!profile) return null
+    const abbreviatedProjectName = getAbbreviatedProjectName(activeProject)
+    const options = visibleProjects.map(
+      project => ({value: project.id, label: project.name})
+    )
```
landonreed, line 38: I added this `options` variable so that we could do `options.find()` to locate the selected value. Arguably, one might want to just transform `activeProject` like you had previously done, but I find it cleaner to reference an already available option.

lib/manager/components/HomeProjectDropdown.js
```diff
@@ -17,20 +18,9 @@ type Props = {
...
-  }
-
-  handleChange = (selectedOption: any, activeProject: any) => {
-    this.setState({selectedOption})
-    activeProject === null
-      ? browserHistory.push(`/home/${selectedOption.value}`)
-      : browserHistory.push(`/home/`)
+export default class HomeProjectDropdown extends Component<Props> {
+  handleChange = (option: ReactSelectOption) => {
+    browserHistory.push(`/home/${option ? option.value : ''}`)
```
landonreed, line 23: Moving the ternary down the chain into the string construction allows us to only invoke `browserHistory.push` once. Ultimately, the same code, but I find this a bit more clear. Another nice way to frame this might be:

```javascript
  handleChange = (option: ReactSelectOption) => {
    const path = option ? '/home/' : `/home/${option.value}`
    browserHistory.push(path)
  }
```

landonreed:
Closing. I think these changes were implemented in #650.

PR #681, opened 2021-05-13
Normalize stop names changes

landonreed:
This suggests a few changes for #649.
1. It removes the Transform.js file that contains a single type. I think the project got into trouble at one point because we had flow type defs scattered across a bunch of files (perf issues).
2. Adds a placeholder method so that type-specific placeholders can be used.
3. Adjusts some labels
4. Disables the substitution rows if performsubstitutions is disabled.

A couple of final questions for you @binh-dam-ibigroup:
1. I'd really like for the onFieldChange behavior to also save the transformation to the back end (to mimic the onFileChange behavior). Is there an easy way to do that? I noticed some issues when I tried to do it due to interference with state/validation changes.
2. My impression of the substitutions is that the boolean `performSubstitutions` is redundant. We should really just have a list of substitutions and if that list is empty, no substitutions will be applied. How big of a tweak would that be?


landonreed:
Actually, I want to check one more thing real quick.

binh-dam-ibigroup:
> 1. I'd really like for the onFieldChange behavior to also save the transformation to the back end (to mimic the onFileChange behavior). Is there an easy way to do that? I noticed some issues when I tried to do it due to interference with state/validation changes.

@landonreed That was my initial approach with the `onChange` event, however there were too much flickering of the blue status bubble while typing. Also, the CSV field for the `ReplaceFileFromStringTransformation` is subject to a `Save` button. Maybe a middleground I have not tried is to save `onBlur` instead, but that would not remove the need for state management in the different components, let me know if you want to try that.

> 2. My impression of the substitutions is that the boolean `performSubstitutions` is redundant. We should really just have a list of substitutions and if that list is empty, no substitutions will be applied. How big of a tweak would that be?

Hmmm why not, I didn't consider that when I converted the capitalization exceptions to a list, that could also saves some vertical space?

lib/manager/components/transform/NormalizeField.js
```diff
@@ -201,15 +199,18 @@ export default class NormalizeField extends Component<TransformProps<NormalizeFi
               checked={capitalize}
               onChange={this._onChangeCapitalize}
             >
-              Capitalize field value
+              Enforce title case (e.g., "Main Street and First Avenue")
```
binh-dam-ibigroup, line 202: "And" would also be capitalized by default without defining an exception.

binh-dam-ibigroup (approved):

PR #692, opened 2021-07-15
Feed transformations: Remove Version Clone trigger, add GTFS+ trigger

binh-dam-ibigroup:
### Checklist

### Description

This PR add a feed transformation trigger for feeds produced in-house with the GTFS+ editor, and removes the trigger for version clone.

Internally, the text for all the trigger is derived using existing string resources instead of using variable names.


miles-grant-ibigroup (approved):
This looks good to me! There are some minor style inconsistencies [such as here](https://github.com/ibi-group/datatools-ui/pull/692/files#diff-9c92995f96e9d7dbdd7e8821b2d960df85b555d8d684b777aec2ccd2a6767602R5-R7) (spaces within import brackets) but I think that is more the linters fault than yours. Should be fixed when/if we lint the entire project with our new lint config!

evansiroky (approved):

lib/common/util/util.js
```diff
@@ -29,14 +31,29 @@ export function versionsSorter (
...
+    case RETRIEVAL_METHODS.PRODUCED_IN_HOUSE:
+      return messages(`${retrievalMethod}.producedInHouse`)
+    case RETRIEVAL_METHODS.PRODUCED_IN_HOUSE_GTFS_PLUS:
+      return messages(`${retrievalMethod}.producedInHouseGtfsPlus`)
+    case RETRIEVAL_METHODS.REGIONAL_MERGE:
+      return messages(`${retrievalMethod}.regionalMerge`)
+    case RETRIEVAL_METHODS.SERVICE_PERIOD_MERGE:
+      return messages(`${retrievalMethod}.servicePeriodMerge`)
+    case RETRIEVAL_METHODS.VERSION_CLONE:
+      return messages(`${retrievalMethod}.versionClone`)
```
landonreed, line 56: If we want to be dynamic with these message keys, we could just convert the snake case with camel case and join with `properties.retrievalMethod`. But if we're defining a case for each of these, why not just have each be a string literal? E.g., `return messages('properties.retrievalMethod.manuallyUploaded')`

landonreed (approved):
Looks good, just one nit about the messages.

lib/common/util/util.js
binh-dam-ibigroup: Thanks for the suggestion. Updated in https://github.com/ibi-group/datatools-ui/pull/692/commits/0bebaf0034ffa431535039c11648ceb651f220d2.

landonreed: Oops, looks like that broke things, @binh-dam-ibigroup.

PR #697, opened 2021-07-29
FeedLabel tweaks

binh-dam-ibigroup:
Some markup cleanup suggestions for the FeedLabel component.

### DIFF
--- a/lib/style.css
+++ b/lib/style.css
@@ -161,13 +161,15 @@ td.feed-source-info {
   width: 100%;
 }
 .feedLabel {
-  align-items: center;
   border-radius: 5px;
   border: 1px solid;
   height: 100%;
   max-width: 100%;
   min-width: 60px;
-  padding: 5px 0 8px 0;
+  overflow: hidden;
+  padding: 5px;
+  text-align: center;
+  text-overflow: ellipsis;
   white-space: nowrap;
   word-break: break-word;
 }
--- a/lib/style.css
+++ b/lib/style.css
@@ -168,6 +168,7 @@ td.feed-source-info {
   max-width: 100%;
   min-width: 60px;
   padding: 5px 0 8px 0;
+  white-space: nowrap;
   word-break: break-word;
 }
 .feedLabel.clickable {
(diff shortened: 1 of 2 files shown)


lib/style.css
```diff
@@ -161,13 +161,16 @@ td.feed-source-info {
...
-  align-items: center;
   border-radius: 5px;
   border: 1px solid;
   height: 100%;
   max-width: 100%;
   min-width: 60px;
-  padding: 5px 0 8px 0;
+  overflow: hidden;
+  padding: 5px;
+  text-align: center;
```
miles-grant-ibigroup, line 171: I think this is what causes the strange unequal padding that sort of pushes the text into the bottom right corner. I don't know of a way to get even alignment using only text rendering. To my knowledge, flexbox is needed.
Screen Shot 2021-07-29 at 1 31 29 PM

miles-grant-ibigroup (commented):
I like this refactoring (especially moving the label content into its own variable), but I don't think we can move the label from a `div` to a `span`. 

It's impossible to center the text correctly within the label if the label itself is the text tag. Even when I force the `span`s to be block elements they don't flow correctly, and the Bootstrap checkbox brings a whole separate can of worms with it.

Screen Shot 2021-07-29 at 1 39 52 PM

Flexbox solves these problems, but requires the extra `div`s that you were trying to remove... I suppose it would be possible to accomplish the same thing using nested `span`s, but you'd have to style them so much you might as well use `div`s. 

If you have a better solution that doesn't use flexbox please let me know because I would be curious, but I haven't found anything else that works with the same consistent results.

lib/style.css
miles-grant-ibigroup: I don't think padding works either, since now the icon would be affected along with the text (since there is no container)

binh-dam-ibigroup: Try the changes from https://github.com/ibi-group/datatools-ui/pull/697/commits/e3dd6d8f8627e4a1dbaaa9e22e0139d79d0d8e9a.

binh-dam-ibigroup: Also, a number of CSS rules have becomed orphaned with my changes, you should remove them if you choose to merge this PR.

miles-grant-ibigroup:
Have added the HTML syntax improvements to #698. Having the checkbox highlight when the label is hovered is really cool!

PR #693, opened 2021-07-20
Add Labels to Frontend

miles-grant-ibigroup:
### Checklist

- [x] Appropriate branch selected _(all PRs must first be merged to `dev` before they can be merged to `master`)_
- [x] Any modified or new methods or classes have helpful JSDoc and code is thoroughly commented
- [x] The description lists all applicable issues this PR seeks to resolve
- [x] The description lists any configuration setting(s) that differ from the default settings
- [x] All tests and CI builds passing

### Description

This makes use of https://github.com/ibi-group/datatools-server/pull/403 to add label functionality for feed sources. Feed sources can have labels applied to them, feed sources can be filtered by label. Labels can be managed in the dashboard. New components have been created to add a new Label, and to assign labels to a feed source.

#### Notes
When assigning labels from the feed sources dashboard, the assigner disappears after each click. This is due to how the existing redux actions are set up. It's possible to create a new action that updates things differently, but I don't want to waste time on that if it isn't needed.

PR #700, opened 2021-08-03
refactor(labels): tweak label editor, label actions

landonreed:
This PR contains a couple of minor tweaks for #693:
- (soft) enforces max length on name/description (might be good to add something server side too?
- adds random color on new label creation
- tweaks on delete message
- fixes URL dirtying on submit

### DIFF
--- a/lib/manager/components/LabelEditor.js
+++ b/lib/manager/components/LabelEditor.js
@@ -18,6 +18,7 @@ import {
 } from 'react-bootstrap'
 import update from 'react-addons-update'
 import { connect } from 'react-redux'
+import tinycolor from 'tinycolor2'
 
 import { createLabel, updateLabel } from '../../manager/actions/labels'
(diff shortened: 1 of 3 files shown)


lib/manager/components/LabelEditor.js
```diff
@@ -52,7 +53,10 @@ class LabelEditor extends React.Component<Props, State> {
         name: true,
         description: true
       },
-      newLabel: {...this.props.label},
+      newLabel: {
+        color: tinycolor.random().toHexString(),
```
miles-grant-ibigroup, line 57: Glad we got a second use of the library!

miles-grant-ibigroup:
Good catch with the `e.preventDefault()` stuff. Will merge and have created https://github.com/ibi-group/datatools-server/pull/403/commits/c4ba8a1f75732645a864edd3b08e745afa1ce1de to do server-side validation

Issue #702, opened 2021-08-03
Add admin-only field to notes


landonreed:
A feature has been requested to make notes on feeds/versions optionally visible to only project admins with a new `adminOnly` field.

PR #703, opened 2021-08-04
Add optional admin-only visibility for notes

landonreed:
### Checklist

### Description

Adds admin only option for a note. re #702

Note viewer is visible at http://localhost:9966/feed/FEED_ID/comments

image

### DIFF
--- a/lib/manager/components/NoteForm.js
+++ b/lib/manager/components/NoteForm.js
@@ -1,0 +1,103 @@
+// @flow
+
+import Icon from '@conveyal/woonerf/components/icon'
+import React, {Component} from 'react'
+import { Button, Checkbox, Col, FormControl, Media, Panel } from 'react-bootstrap'
+
+import {getComponentMessages} from '../../common/util/config'
+import { getProfileLink } from '../../common/util/util'
+import type {Feed} from '../../types'
+import type {ManagerUserState} from '../../types/reducers'
+
+type Props = {
+  feedSource: Feed,
+  newNotePosted: ({body: string}) => void,
+  stacked?: boolean,
+  user: ManagerUserState
+}
+
+type State = {
+  adminOnly: boolean,
+  body: string
+}
+
+const DEFAULT_STATE = {
+  adminOnly: false,
+  body: ''
+}
+
+export default class NoteForm extends Component<Props, State> {
+  messages = getComponentMessages('NoteForm')
+  state = DEFAULT_STATE
+
+  _onChange = ({target}: {target: HTMLInputElement}) =>
+    this.setState({[target.name]: target.type === 'checkbox' ? target.checked : target.value})
+
+  _onClickPublish = () => {
+    this.props.newNotePosted(this.state)
+    this.setState(DEFAULT_STATE)
+  }
+
+  render () {
+    const {
+      feedSource,
+      stacked,
+      user
+    } = this.props
+    const {Body, Heading, Left} = Media
+    const {adminOnly, body} = this.state
+    const {profile} = user
+    if (!profile) {
+      console.warn('Could not locate user profile', user)
+      return null
+    }
+    const isProjectAdmin = user.permissions && user.permissions.isProjectAdmin(
+      feedSource.projectId, feedSource.organizationId
+    )
+    const userLink = user
+      ? <a href={getProfileLink(profile.email)}>{profile.email}</a>
+      : 'Unknown user'
+    return (
+      <Col xs={12} sm={stacked ? 12 : 4} md={stacked ? 12 : 6}>
+        <h3>{this.messages('postComment')}</h3>
+        <Media>
+          <Left>
+            <img
+              alt={profile.email}
+              width={64}
+              height={64}
+              src={user ? profile.picture : ''} />
+          </Left>
+          <Body>
+            <Panel
+              className='comment-panel'
+              header={<Heading>{userLink}</Heading>}>
+              <FormControl
+                ref='newNoteBody'
+                componentClass='textarea'
+                name='body'
+                value={body}
+                onChange={this._onChange} />
+              <Button
+                className='pull-right'
+                style={{marginTop: '10px'}}
+                disabled={body === ''}
+                onClick={this._onClickPublish}>
+                {this.messages('new')}
+              </Button>
+              {isProjectAdmin &&
+                <Checkbox
+                  name='adminOnly'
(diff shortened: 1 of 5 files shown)


lib/manager/components/NoteForm.js
```diff
@@ -0,0 +1,103 @@
...
+    const {profile} = user
+    if (!profile) {
+      console.warn('Could not locate user profile', user)
+      return null
+    }
+    const isProjectAdmin = user.permissions && user.permissions.isProjectAdmin(
+      feedSource.projectId, feedSource.organizationId
+    )
+    const userLink = user
+      ? <a href={getProfileLink(profile.email)}>{profile.email}</a>
```
miles-grant-ibigroup, line 58: This generated a link that sent me to a 404 page. Might be worth refactoring `getProfileLink` to return null if the link won't be valid and then render a span instead?

miles-grant-ibigroup (changes requested):
Looks good! Just a few really small UI bugs

lib/manager/components/NoteForm.js
```diff
@@ -0,0 +1,103 @@
...
+              header={<Heading>{userLink}</Heading>}>
+              <FormControl
+                ref='newNoteBody'
+                componentClass='textarea'
+                name='body'
+                value={body}
+                onChange={this._onChange} />
+              <Button
+                className='pull-right'
+                style={{marginTop: '10px'}}
```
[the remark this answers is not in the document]
landonreed: Thanks! Fixed in e52fced352d423d646aa501a60e155abb707cdee

lib/manager/components/NoteForm.js
landonreed: **Good catch, but tldr I think this is out of scope.**

This could get complicated, so I think's it's probably out of scope (this isn't really new code, just pulled out of NotesViewer). Also, rendering just a span will not allow the user to click through and potentially create a profile image if they're so inclined (though I doubt they will care).

In any case, the profile link, whether valid or invalid, triggers a redirect to either the valid gravatar URL (200) or the 404 page.

Here's my profile link: https://en.gravatar.com/466d1e90516eab20031b225df44decdd

Inspect the requests to see it in action. Handling all of this in our codebase doesn't sound like a great idea haha. There might be a more elegant way to go about it, but I'm not sure it's worth the time.

lib/manager/components/NoteForm.js
```diff
@@ -0,0 +1,103 @@
...
+          </Left>
+          <Body>
+            <Panel
+              className='comment-panel'
+              header={<Heading>{userLink}</Heading>}>
+              <FormControl
+                ref='newNoteBody'
+                componentClass='textarea'
+                name='body'
+                value={body}
```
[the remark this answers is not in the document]
landonreed: Woops, I think that `none` needs quotes, right?

lib/manager/components/NoteForm.js
```diff
@@ -0,0 +1,103 @@
...
+    return (
+      <Col xs={12} sm={stacked ? 12 : 4} md={stacked ? 12 : 6}>
+        <h3>{this.messages('postComment')}</h3>
+        <Media>
+          <Left>
+            <img
+              alt={profile.email}
+              width={64}
+              height={64}
+              src={user ? profile.picture : ''} />
```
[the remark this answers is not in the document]
landonreed: Fixed in abda8672a1904abd02e05cc6c8d4edd6370ae9a1

lib/manager/components/NoteForm.js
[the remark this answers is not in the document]
landonreed: Fixed in abda8672a1904abd02e05cc6c8d4edd6370ae9a1

miles-grant-ibigroup (approved):
Good point with the Gravatar stuff. Didn't realize it was that deep a rabbit hole. Otherwise looks good! I'm surprised they didn't ask to remove/edit/archive comments but maybe that's a task for later.

lib/manager/components/NoteForm.js
```diff
@@ -0,0 +1,110 @@
...
+              }
+            >
+              <FormControl
+                componentClass='textarea'
+                name='body'
+                ref='newNoteBody'
+                style={{resize: 'none'}}
+                onChange={this._onChange}
+                value={body} />
+              <Button
```
binh-dam-ibigroup, line 86: Should this button be the `submit` button of inside a `<form>`?

lib/manager/components/NoteForm.js
```diff
@@ -0,0 +1,110 @@
...
+  adminOnly: boolean,
+  body: string
+}
+
+const DEFAULT_STATE = {
+  adminOnly: false,
+  body: ''
+}
+
+export default class NoteForm extends Component<Props, State> {
```
binh-dam-ibigroup, line 29: Add JSDoc

lib/manager/components/NoteForm.js
```diff
@@ -0,0 +1,110 @@
...
+              header={
+                <Heading>
+                  <a href={getProfileLink(profile.email)}>{profile.email}</a>
+                </Heading>
+              }
+            >
+              <FormControl
+                componentClass='textarea'
+                name='body'
+                ref='newNoteBody'
```
binh-dam-ibigroup, line 82: If the ref used? (I don't see another instance of `newNoteBody`)

lib/manager/components/NotesViewer.js
```diff
@@ -107,6 +86,13 @@ export default class NotesViewer extends Component<Props, State> {
                           <small title={noteDate.format('h:MMa, MMM. DD YYYY')}>
                             commented {noteDate.fromNow()}
                           </small>
+                          {note.adminOnly &&
+                            <span
+                              title={this.messages('adminOnly')}
+                              className='pull-right'>
+                              <Icon type='lock' />
+                            </span>
+                          }
```
binh-dam-ibigroup, line 95: This content should be first under the `<Heading>` tag to avoid spilling on narrow displays.
image

binh-dam-ibigroup: Do we need the extra `<span>` that wraps around the padlock icon?

binh-dam-ibigroup (approved):
LGTM after fixing a few formatting comments.

lib/manager/components/NotesViewer.js
landonreed: Both great catches.

PR #765, opened 2022-02-27
Display null continuous_pickup as 1 (no continous pickup)

philip-cline:
### Checklist

### Description

Empty values of continuous_pickup and drop_off are misinterpreted by the front end as “0 - Available”. The correct value is 1 - Not Available.

### DIFF
--- a/lib/editor/components/pattern/PatternStopCard.js
+++ b/lib/editor/components/pattern/PatternStopCard.js
@@ -424,7 +424,7 @@ class PatternStopContents extends Component<Props, State> {
               selectType='continuousPickup'
               shouldHaveDisabledOption
               title='Indicates whether a rider can board the transit vehicle anywhere along the vehicle’s travel path.'
-              value={patternStop.continuousPickup || ''}
+              value={patternStop.continuousPickup || 1}
             />
           </Col>
           <Col xs={6}>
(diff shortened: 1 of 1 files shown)


binh-dam-ibigroup:
> ~@philip-cline Should this change apply to the route editor too?~

Scratch that comment, that attribute is optional in `routes.txt`, but is there a way to make those fields optional in the pattern editor?

philip-cline:
Making the fields optional would also apply then to `pickup_type` and `drop_off_type` which are also optional in the spec. I can take a stab at updating these & also making sure behaviour around the routes.txt continuous pickup and dropoff is working correctly

philip-cline:
> Another (probably important) thing to consider is the spec allows the stop times with unspecified continuous pickup/drop off to "inherit" those the corresponding settings of the parent route (see spec link above).

Yep definitely, although does this mean we would want to display stop_time level continuous_pickup & drop_off as the inherited value, until is overridden? This seems closest to user intention, but would take me more time to implement (so maybe get the correct display out there in this PR, & the inheritance and override in a following PR).

binh-dam-ibigroup:
> > Another (probably important) thing to consider is the spec allows the stop times with unspecified continuous pickup/drop off to "inherit" those the corresponding settings of the parent route (see spec link above).
> 
> Yep definitely, although does this mean we would want to display stop_time level continuous_pickup & drop_off as the inherited value, until is overridden? This seems closest to user intention, but would take me more time to implement (so maybe get the correct display out there in this PR, & the inheritance and override in a following PR).

I am fine with doing that in a separate PR.

miles-grant-ibigroup (approved):
Patch works great for me thanks!

PR #767, opened 2022-03-04
Allow language code to match code with region

philip-cline:
### Checklist

### Description

This PR adds support for matching a language code without a region to the specific ID in our language files (i.e. "pl" is matched because it is included in "pl-PL"). This should hopefully allow easier use of internationalization in datatools.

### DIFF
--- a/lib/common/util/config.js
+++ b/lib/common/util/config.js
@@ -121,8 +121,9 @@ function initializeConfig () {
   const languageId = window.localStorage.getItem('lang')
     ? window.localStorage.getItem('lang')
     : navigator.language
-  const active = languages.find(l => languageId.includes(l._id)) ||
-    languages.find(l => l._id === 'en-US')
+  const active = languages.find(
+    l => l._id === languageId || languageId.startsWith(l._id)
+  ) || languages.find(l => l._id === 'en-US')
   if (!active) throw new Error('Language file is misconfigured!')
   // is an array containing all the matching modules
   config.messages = {
--- a/lib/common/util/config.js
+++ b/lib/common/util/config.js
@@ -121,7 +121,7 @@ function initializeConfig () {
   const languageId = window.localStorage.getItem('lang')
     ? window.localStorage.getItem('lang')
     : navigator.language
-  const active = languages.find(l => l._id.includes(languageId)) ||
+  const active = languages.find(l => languageId.includes(l._id)) ||
     languages.find(l => l._id === 'en-US')
   if (!active) throw new Error('Language file is misconfigured!')
   // is an array containing all the matching modules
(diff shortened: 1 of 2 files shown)


lib/common/util/config.js
```diff
@@ -121,7 +121,7 @@ function initializeConfig () {
   const languageId = window.localStorage.getItem('lang')
     ? window.localStorage.getItem('lang')
     : navigator.language
-  const active = languages.find(l => l._id === languageId) ||
+  const active = languages.find(l => l._id.includes(languageId)) ||
```
miles-grant-ibigroup, line 124:
```suggestion
  const active = languages.find(l => languageId.includes(l._id) ||
```
Could we swap it around like this? That way we can set our language string in the language file to just `pl` and then a browser string of `pl` or `pl-PL` will match. right?

miles-grant-ibigroup (changes requested):
Please let me know what you think of my suggestion to flip around how we do the check... might match in more situations?

lib/common/util/config.js
miles-grant-ibigroup: Although if it works for polish as is, then I'd say we can leave it as is!

philip-cline: Yeah I like that, updated to this in https://github.com/ibi-group/datatools-ui/pull/767/commits/f0c2107f1fe807427265d93968cb6339a6cc388f

miles-grant-ibigroup (approved):
Tried it locally and swapped out the English and Polish ids and it seems to have worked. Thanks for the changes!

lib/common/util/config.js
```diff
@@ -121,7 +121,7 @@ function initializeConfig () {
   const languageId = window.localStorage.getItem('lang')
     ? window.localStorage.getItem('lang')
     : navigator.language
-  const active = languages.find(l => l._id === languageId) ||
+  const active = languages.find(l => languageId.includes(l._id)) ||
```
binh-dam-ibigroup, line 124: I would say `l => l._id === languageId || languageId.startsWith(l._id)` to match the exact locale if it exists. 

But definitely at least: `l => languageId.startsWith(l._id)`.

binh-dam-ibigroup (approved):
Conditional approval, see the noted change in the language check.

lib/common/util/config.js
philip-cline: Definitely agree with this, updated to the first option in https://github.com/ibi-group/datatools-ui/pull/767/commits/a6e06b228d58580bd945bda5ec72adc12c5e0890

binh-dam-ibigroup (approved):
Approval stands.

PR #772, opened 2022-03-16
Fix timetable previous stop time checks from checking text columns.

philip-cline:
### Checklist

### Description

Currently, naming a trip ID (or filling any other text field) with a number greater than the start time of the first arrival time produces a validation error in the UI since validation takes place for _all_ columns left of the stop time. This PR adds a condition to only check columns that are not of type 'TEXT'. 

The PR also adds VS code `settings.json` to the `.gitignore` file, which I hope won't be controversial. 

Issue screenshot: 
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,4 +30,4 @@ env.yml-original
 scripts/*client.json
 
 # Vs code settings
-.vscode
+.vscode/
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,6 @@ env.yml-original
 .env
 !configurations/test/env.yml
 scripts/*client.json
+
+# Vs code settings
+.vscode
--- a/lib/editor/selectors/timetable.js
+++ b/lib/editor/selectors/timetable.js
@@ -150,7 +150,7 @@ const isCellValueInvalid = (
       let previousValue = null
       let previousIndex = colIndex - 1
       // Find previous stop time
-      while (previousValue === null && previousIndex >= 0 && columns[previousIndex].type !== 'TEXT') {
+      while (previousValue === null && previousIndex >= 0 && (columns[previousIndex].type === 'ARRIVAL_TIME' || columns[previousIndex].type === 'DEPARTURE_TIME')) {
         const previousCol = columns[previousIndex]
         previousValue = previousCol && row && objectPath.get(row, previousCol.key)
         previousIndex--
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -28,3 +28,6 @@ env.yml-original
 .env
 !configurations/test/env.yml
 scripts/*client.json
+
+# Vs code settings
+.vscode
```
miles-grant-ibigroup, line 33: Super nit I'd add a trailing slash here to make it clear it's a folder!

miles-grant-ibigroup (approved):
I can't believe this was the fix... well done figuring it out!

.gitignore
philip-cline: Super nit could be a great marvel character, you may want to trademark that

philip-cline: Fixed in https://github.com/ibi-group/datatools-ui/pull/772/commits/da2ba045945c8b64d06d0c3f207b8d38b89759f2

lib/editor/selectors/timetable.js
```diff
@@ -152,7 +150,7 @@ const isCellValueInvalid = (
       let previousValue = null
       let previousIndex = colIndex - 1
       // Find previous stop time
-      while (previousValue === null && previousIndex >= 0) {
+      while (previousValue === null && previousIndex >= 0 && columns[previousIndex].type !== 'TEXT') {
```
binh-dam-ibigroup, line 153: Wondering whether we should check whether the previous column index is the last column index for non-stop-time trip data?

binh-dam-ibigroup (approved):
It would be nice of the validation ignored the first few columns instead, but as written it is fine.

lib/editor/selectors/timetable.js
philip-cline: That's a good idea, I changed the condition to only be if the type is `ARRIVAL_TIME` or `DEPARTURE_TIME` in https://github.com/ibi-group/datatools-ui/pull/772/commits/896ebe0fcff0da1e7ffe91adf90d7915fdd232fe

philip-cline: if you'd prefer I guess we could find the index of the first stop time column and check based on that, but this seems simpler

binh-dam-ibigroup: What you did using `ARRIVAL_TIME` or `DEPARTURE_TIME` is good too!

PR #777, opened 2022-04-05
Update config for mkdocs buiid.

binh-dam-ibigroup:
### Checklist

- [x] Appropriate branch selected _(all PRs must first be merged to `dev` before they can be merged to `master`)_
- [na] Any modified or new methods or classes have helpful JSDoc and code is thoroughly commented
- [na] The description lists all applicable issues this PR seeks to resolve
- [na] The description lists any configuration setting(s) that differ from the default settings
- [x] All tests and CI builds passing

### Description

This PR fixes mkdocs.yml to match the config structure expected by recent versions of mkdocs.

PR #778, opened 2022-04-07
fix(timetable): Allow offsetting by 24hrs or more

philip-cline:
### Checklist

### Description

Currently in datatools, offsetting by a value of 24 hours or more in the timetable editor will result in an offset of that value modulus 86399 (1 second if offset is 24 hours). However, transit agencies will want to offset by 24 hours or greater for a trip that begins past midnight on the same service day as other trips. They currently use a workaround where the same result can be achieved by offsetting by 12h twice. This PR enables offsetting by 24 hours or more, to bring DT behaviour in line with service day offsets. 

**Tests will fail until https://github.com/ibi-group/datatools-ui/pull/777 is merged

### DIFF
--- a/lib/editor/reducers/timetable.js
+++ b/lib/editor/reducers/timetable.js
@@ -7,7 +7,6 @@ import clone from 'lodash/cloneDeep'
 import { sortAndFilterTrips } from '../util'
 import { isTimeFormat } from '../util/timetable'
 import {resequenceStops} from '../util/map'
-
 import type {Action} from '../../types/actions'
 import type {TimetableState} from '../../types/reducers'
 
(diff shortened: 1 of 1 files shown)


lib/editor/reducers/timetable.js
```diff
@@ -98,7 +97,8 @@ const timetable = (state: TimetableState = defaultState, action: Action): Timeta
           const path = `${action.payload.rowIndexes[i]}.${col.key}`
           if (isTimeFormat(col.type)) {
             const currentVal = objectPath.get(trips, path)
-            const value = currentVal + (action.payload.offset % 86399) // ensure seconds does not exceed 24 hours
+            // Arrival time can be > 24 hours for service day, so do not block offsetting by 24 hours.
+            const value = currentVal + action.payload.offset
```
binh-dam-ibigroup, line 101: If there are blank cells in the row that is being offset, the blank cells will be filled with the time with the modulo applied if the applicable stop time is after midnight, and a validation error is shown for these cells because the time in these cells is before the time in the cell referenced in this code. Do we want to remove that behavior for those cells too?

binh-dam-ibigroup (commented):
See comment thread at https://github.com/ibi-group/datatools-ui/pull/778/files#r846143817

lib/editor/reducers/timetable.js
philip-cline: I think the better behaviour is to ignore cells that are initially blank

binh-dam-ibigroup (approved):

daniel-heppner-ibigroup (approved):
LGTM

PR #784, opened 2022-04-19
style(Job Monitor Popover): Fix exception overflow, allow scrolling in CSS

philip-cline:
### Checklist

### Description

Previously, Datatools backend exceptions (if they occurred) would appear in the Job Monitor Popover overflowing the component's width (fake text in exception used for demonstration)
image

In this PR, scrolling is added to prevent this behaviour: 
image

### DIFF
--- a/lib/style.css
+++ b/lib/style.css
@@ -996,3 +996,8 @@ h4.line:after {
 .route-type-select .toggle.expanded::after {
   content: "\f068";
 }
+
+/* Datatools backend exception in popover overflows the container */
+.popover-content {
+  overflow-x: scroll;
+}


lib/style.css
```diff
@@ -996,3 +996,8 @@ h4.line:after {
 .route-type-select .toggle.expanded::after {
   content: "\f068";
 }
+
+/* Datatools backend exception in popover overflows the container */
+.popover-content {
+  overflow-x: scroll;
```
miles-grant-ibigroup, line 1002: Is there any way to increase the size of the popup? Otherwise I guess scrolling is alright as well. I think the scrollbar will appear by default right

miles-grant-ibigroup (approved):
Looks good

lib/style.css
philip-cline: The size of the popup could be expanded, but personally I find the scrollbar to be the best of both worlds - cleaner display while maintaining option for the user (generally us) to investigate details.

daniel-heppner-ibigroup (approved):

PR #786, opened 2022-04-26
Fix undeletable patternstops

philip-cline:
### Checklist

### Description

This PR resolves an issue where adding stops after a straight, two stop segment produced an undeletable patternStop. This behaviour can be seen [here](https://drive.google.com/file/d/1cs7zLb5BFY1040rpH3aahgokqcQs3Ahj/view?usp=sharing). The cause of the behaviour appears to have been a typo in the definition of `hasShapePoints` (other functions use the definition this PR introduces)


philip-cline:
This fix seems suspiciously too simple, so definitely looking for a strong sanity check on this one.

miles-grant-ibigroup (approved):
I've tried to find the origin of this 2 but the only explanation I can come up with is typo. Good job finding the source of the bug!

daniel-heppner-ibigroup (approved):
Wow, insane bug hunting skills! Nice one. It *could* be refactored to use the `?` operator, but it doesn't really matter.

PR #812, opened 2022-06-24
Add location groups

miles-grant-ibigroup:
### Checklist

### Description

This PR adds a location group editor. Blocked waiting for server-side changes that allow these changes to be exported properly.

Blocked by https://github.com/conveyal/gtfs-lib/pull/353

A future PR will handle adding stops/locations to a location group from the map.

### TODO
- [x] Adding location groups to patterns
- [x] Location Group editor could look a little nicer


lib/editor/actions/map/stopStrategies.js
```diff
@@ -231,15 +231,17 @@ export function addStopToPattern (pattern: Pattern, stop: GtfsStop | GtfsLocatio
     const {avoidMotorways, followStreets} = editSettings.present
     const {
       patternLocations: currentPatternLocations,
+      patternLocationGroups: currentPatternLocationGroups,
```
binh-dam-ibigroup, line 233: Sort attributes.

lib/editor/actions/tripPattern.js
```diff
@@ -12,10 +12,10 @@ import {fetchGTFSEntities, receiveGTFSEntities} from '../../manager/actions/vers
 import {getEditorNamespace} from '../util/gtfs'
 import {resequenceStops, resequenceShapePoints} from '../util/map'
 import {entityIsNew} from '../util/objects'
-import type {ControlPoint, Pattern, PatternLocation, PatternStop} from '../../types'
+import type {ControlPoint, Pattern, PatternLocationGroup, PatternLocation, PatternStop} from '../../types'
```
binh-dam-ibigroup: Sort imports (2 instances)

lib/editor/actions/tripPattern.js
```diff
@@ -79,28 +79,31 @@ export function normalizeStopTimes (patternId: number, beginStopSequence: number
  */
 export function updatePatternStops (
   pattern: Pattern,
-  patternHalts: Array<PatternStop | PatternLocation>
+  patternHalts: Array<PatternStop | PatternLocation | PatternLocationGroup>
 ) {
   return function (dispatch: dispatchFn, getState: getStateFn) {
     const sortedPatternHalts = patternHalts.map(resequenceStops)
-    const {stops, locations} = sortedPatternHalts.reduce(
+    const {stops, locations, locationGroups} = sortedPatternHalts.reduce(
```
binh-dam-ibigroup: Sort attributes (multiple instances)

lib/editor/components/EditorInput.js
```diff
@@ -15,14 +15,15 @@ import {doesNotExist} from '../util/validation'
 import TimezoneSelect from '../../common/components/TimezoneSelect'
 import LanguageSelect from '../../common/components/LanguageSelect'
 import toSentenceCase from '../../common/util/to-sentence-case'
-import type {Entity, Feed, GtfsSpecField, GtfsAgency, GtfsStop} from '../../types' // Add LocationGroupOption
+import type {Entity, Feed, GtfsSpecField, GtfsAgency, GtfsLocation, GtfsStop} from '../../types' // Add LocationGroupOption
```
binh-dam-ibigroup: Sort types and delete comment

lib/editor/components/EditorInput.js
```diff
@@ -400,6 +400,22 @@ export default class EditorInput extends React.Component<Props> {
...
+        const stops: Array<GtfsStop> = [...getTableById(tableData, 'stop')]
+        const locations: Array<GtfsLocation> = [...getTableById(tableData, 'location')]
+
+        return (
+          <FormGroup {...formProps}>
+            {basicLabel}
+            <LocationGroupsEditor
+              currentValue={currentValue}
+              stops={stops}
+              locations={locations}
```
binh-dam-ibigroup, line 411: Sort props

lib/editor/components/EditorInput.js
```diff
@@ -400,6 +400,22 @@ export default class EditorInput extends React.Component<Props> {
           </FormGroup>
         )
       }
+      case 'GTFS_STOP_OR_LOCATION_LIST': {
+        const stops: Array<GtfsStop> = [...getTableById(tableData, 'stop')]
+        const locations: Array<GtfsLocation> = [...getTableById(tableData, 'location')]
```
binh-dam-ibigroup: Are these arrays expected to be modified in this component or as passed in the children (despite the fact that other cases above are written similar.y)? Otherwise the result of getTableById could just be passed instead.

lib/editor/components/LocationGroupsEditor.js
```diff
@@ -0,0 +1,86 @@
...
+import VirtualizedEntitySelect from './VirtualizedEntitySelect'
+
+type Props = {
+  currentValue: Array<string> | string,
+  locations: Array<GtfsLocation>,
+  processFieldChange: (val: any) => void,
+  stops: Array<GtfsStop>
+}
+
+const LocationGroupsEditor = ({ currentValue, stops, locations, processFieldChange }: Props) => {
```
binh-dam-ibigroup: Sort props

lib/editor/components/LocationGroupsEditor.js
```diff
@@ -0,0 +1,86 @@
...
+  const [dropdownShowing, setDropdownShowing] = useState(false)
+  const getStationOrLocationName =
+    (haltId) => {
+      const entity =
+        stops.find((stop) => stop.stop_id === haltId) ||
+        locations.find((location) => location.location_id === haltId)
+      return entity ? entity.stop_name : 'mystery'
+    }
+  if (!currentValue) currentValue = []
```
binh-dam-ibigroup: Move this to function declaration.

lib/editor/components/LocationGroupsEditor.js
```diff
@@ -0,0 +1,86 @@
...
+    if (typeof currentValue === 'string') return
+
+    processFieldChange(
+      currentValue.filter(id => id !== haltId)
+    )
+  }
+
+  return (
+    <>
+      <div style={{display: 'grid', gap: 5, gridTemplateColumns: '6fr 1fr', marginBottom: 5}}>
```
binh-dam-ibigroup: Sort CSS

lib/editor/components/LocationGroupsEditor.js
```diff
@@ -0,0 +1,86 @@
...
+    (haltId) => {
+      const entity =
+        stops.find((stop) => stop.stop_id === haltId) ||
+        locations.find((location) => location.location_id === haltId)
+      return entity ? entity.stop_name : 'mystery'
+    }
+  if (!currentValue) currentValue = []
+
+  const deleteHalt = (haltId) => {
+    if (typeof currentValue === 'string') return
```
binh-dam-ibigroup, line 32: Extract variable and reuse.

lib/editor/components/LocationGroupsEditor.js
binh-dam-ibigroup: This is should probably be an `<ul>`.

lib/editor/components/LocationGroupsEditor.js
```diff
@@ -0,0 +1,86 @@
...
+      currentValue.filter(id => id !== haltId)
+    )
+  }
+
+  return (
+    <>
+      <div style={{display: 'grid', gap: 5, gridTemplateColumns: '6fr 1fr', marginBottom: 5}}>
+        {typeof currentValue !== 'string' &&
+          currentValue.map((l) => (
+            <React.Fragment key={l}>
```
binh-dam-ibigroup, line 49: This should probably be an `<li>`.

lib/editor/components/LocationGroupsEditor.js
```diff
@@ -0,0 +1,86 @@
...
+      >
+        <Icon type='plus' /> Add stop or location by name
+      </Button>}
+      {dropdownShowing && <VirtualizedEntitySelect
+        component={'stop or location'}
+        entityKey={'stop_or_location_id'}
+        entities={[
+          ...stops,
+          ...locations.map((l) => {
+            return { ...l, id: uuidv4() }
```
binh-dam-ibigroup: What does `id` do for location entries?

lib/editor/components/map/EditorMap.js
```diff
@@ -378,6 +382,22 @@ export default class EditorMap extends Component<Props, State> {
...
+      case 'locationgroup':
+        if (!activeEntity) return null
+        const castedLocationGroup = ((activeEntity: any): GtfsLocation)
+
+        const idsToShow = castedLocationGroup.location_id
+        if (!idsToShow || idsToShow.length === 0) return null
+        const locationGroupLocations = locations.filter(location => idsToShow.includes(location.location_id))
+        const locationGroupStops = stops.filter(stop => idsToShow.includes(stop.stop_id))
+
+        return <FeatureGroup>
```
binh-dam-ibigroup: Surround tag with parentheses.

lib/editor/components/map/EditorMap.js
```diff
@@ -378,6 +382,22 @@ export default class EditorMap extends Component<Props, State> {
...
+          <StopsLayer
+            stops={locationGroupStops}
+            mapState={mapState}
```
binh-dam-ibigroup: Sort props

lib/editor/components/LocationGroupsEditor.js
binh-dam-ibigroup: Inline `return`.

lib/editor/components/map/PatternLocationMarker.js
```diff
@@ -86,6 +87,8 @@ export default class PatternLocationMarker extends Component<Props> {
 
   render () {
     const {active, index, location, patternLocation} = this.props
+    if (!location) return null
```
binh-dam-ibigroup, line 92: Hum, not sure I understand this at first sight.

lib/editor/components/map/PatternLocationMarker.js
```diff
@@ -110,44 +113,48 @@ export default class PatternLocationMarker extends Component<Props> {
...
-                    <FormGroup
-                      controlId='defaultTimeInLocation'>
-                      <ControlLabel>Time in location</ControlLabel>
-                      <MinuteSecondInput
-                        seconds={patternLocation.flexDefaultZoneTime}
-                        onChange={this._onChangeTimeInLocation} />
-                    </FormGroup>
-                  </Col>
-                </Row>
+                {patternHaltIsLocation(patternLocation) && <React.Fragment>
```
binh-dam-ibigroup: Should something else be shown if this is not a location?

binh-dam-ibigroup: Sort props, and I would reformat the tags here so that opening and closing brackets are on the same indent (same with `PatternStopMarker`).

lib/editor/components/map/PatternStopsLayer.js
```diff
@@ -18,34 +18,41 @@ import PatternLocationMarker from './PatternLocationMarker'
...
+  locations: Array<GtfsLocation>,
   patternEdited: boolean,
   patternSegment: number,
   patternStop: {id: any, index: number},
   removeStopFromPattern: typeof stopStrategiesActions.removeStopFromPattern,
   saveActiveGtfsEntity: typeof activeActions.saveActiveGtfsEntity,
   setActiveEntity: typeof activeActions.setActiveEntity,
   setActiveStop: typeof tripPatternActions.setActiveStop,
-  updatePatternStops: typeof tripPatternActions.updatePatternStops
+
```
binh-dam-ibigroup: Delete blank line

lib/editor/components/pattern/PatternStopButtons.js
```diff
@@ -43,6 +43,11 @@ export default class PatternStopButtons extends Component<Props> {
     if (stop.hasOwnProperty('stop_id')) {
       setActiveEntity(feedSource.id, 'stop', stop)
     }
+    if (stop.hasOwnProperty('location_group_id')) {
+      setActiveEntity(feedSource.id, 'locationgroup', stop)
+      // Location groups also have a location_id, so we need to drop out here
+      return
```
binh-dam-ibigroup: Replace this with `else if` blocks.

lib/editor/components/pattern/PatternStopCard.js
```diff
@@ -22,10 +22,10 @@ import * as tripPatternActions from '../../actions/tripPattern'
 import { getEntityName, getAbbreviatedStopName, getTableById } from '../../util/gtfs'
 import MinuteSecondInput from '../MinuteSecondInput'
 import { getGtfsSpecField } from '../../../common/util/config'
-import type {Feed, Pattern, PatternLocation, PatternStop} from '../../../types'
+import type {Feed, Pattern, PatternLocation, PatternStop, PatternLocationGroup} from '../../../types'
```
binh-dam-ibigroup: Sort imports

lib/editor/components/pattern/PatternStopCard.js
```diff
@@ -416,6 +416,8 @@ class PatternStopContents extends Component<Props, State> {
   // eslint-disable-next-line complexity
   render () {
     const {active, activePattern, feedSource, patternEdited, patternStop, tables} = this.props
+
+    if (!this.state) return null
```
binh-dam-ibigroup: This looks bad. Better initialize the state field in the class constructor and check individual state fields instead.

lib/editor/components/pattern/PatternStopContainer.js
```diff
@@ -157,12 +159,18 @@ class PatternStopContainer extends Component<Props, State> {
           }
           // Since stops and locations are unified, the stop could also refer to a location
           // IDs exist in one namespace as per the spac
-          if (!stop) {
+          if (card.hasOwnProperty('locationId') && !card.hasOwnProperty('locationGroupId')) {
```
binh-dam-ibigroup: Given the number of times these checks are made, I'm starting to think we need a some util function that makes the distinction (or a function that returns an object of the correct type so you don't have to write all of these flow waivers).

lib/editor/components/pattern/PatternStopsPanel.js
```diff
@@ -195,7 +199,7 @@ export default class PatternStopsPanel extends Component<Props, State> {
             ? <div className='pattern-stop-card'>
               <VirtualizedEntitySelect
                 component={'stop or location'}
-                entities={[...stops, ...locations.map(l => { return {...l, id: uuidv4()} })]}
+                entities={[...stops, ...locations.map(l => { return {...l, id: uuidv4()} }), ...locationGroups.map(lg => { return {...lg, id: uuidv4()} })]}
```
binh-dam-ibigroup, line 203: Add a comment on what these always changing uuids are for.

lib/editor/selectors/timetable.js
```diff
@@ -64,24 +64,24 @@ export const getTimetableColumns = createSelector(
...
-          // $FlowFixMe FLEX TODO: improve this to avoid double ternary
-          const stopName = stop ? stop.stop_name : ps.stopId ? ps.stopId : ps.locationId
+          // $FlowFixMe FLEX TODO: improve this to avoid quadruple ternary
+          const stopName = stop ? stop.stop_name : ps.stopId ? ps.stopId : (ps.locationId ? ps.locationId : ps.locationGroupId)
           const stopId = stop && stop.stop_code
             ? stop.stop_code
             : stop && stop.stop_id
               ? stop.stop_id
-              : stop && stop.location_id ? stop.location_id : ''
+              : stop && stop.location_group_id ? stop.location_group_id : stop && stop.location_id ? stop.location_id : ''
```
binh-dam-ibigroup, line 80: Holy moly it feels like playing Mario!

lib/editor/util/ui.js
```diff
@@ -85,5 +85,14 @@ export const GTFS_ICONS = [
     title: 'Edit Booking Rules',
     label: 'Booking Rules',
     flexOnly: true
+  },
+  {
+    id: 'locationgroup',
```
binh-dam-ibigroup: Sort attributes

binh-dam-ibigroup (commented):
First round of comments. I think what is standing out at first reading is the need of util functions to distinguish/return a desired object type for a stop, location, and location group, so that the code is less verbose.

lib/editor/actions/tripPattern.js
```diff
@@ -79,28 +79,31 @@ export function normalizeStopTimes (patternId: number, beginStopSequence: number
  */
 export function updatePatternStops (
   pattern: Pattern,
-  patternHalts: Array<PatternStop | PatternLocation>
+  patternHalts: Array<PatternStop | PatternLocation | PatternLocationGroup>
```
binh-dam-ibigroup: Create an alias type for `PatternStop | PatternLocation | PatternLocationGroup`. This type combo shows up in many files changed by this PR.

lib/editor/components/EditorInput.js
miles-grant-ibigroup: Good catch thank you! This came from lots of copy/pasting

lib/editor/components/LocationGroupsEditor.js
miles-grant-ibigroup: Good catch thanks!

lib/editor/components/LocationGroupsEditor.js
miles-grant-ibigroup: This used to be a list, but that conflicted with the grid styles.

lib/editor/components/LocationGroupsEditor.js
miles-grant-ibigroup: Extracting the typeof results in needing FlowFixMe's everywhere the variable is used. I'm happy to do the extraction, but am not sure which evil we dislike less haha

lib/editor/components/LocationGroupsEditor.js
miles-grant-ibigroup: Agreed, see previous note

lib/editor/components/pattern/PatternStopCard.js
miles-grant-ibigroup: This is residue from when this was being developed. It looks like it can be safely removed.

lib/editor/components/pattern/PatternStopContainer.js
miles-grant-ibigroup: There is already `patternHaltIsLocation` and `patternHaltIsLocationGroup` which I've now used here.

lib/editor/selectors/timetable.js
miles-grant-ibigroup: 🥲

lib/editor/util/ui.js
miles-grant-ibigroup: I think the order of fields in all of these objects was deliberately chosen to make the large list more readable.

lib/editor/components/LocationGroupsEditor.js
```diff
@@ -0,0 +1,92 @@
...
+  return (
+    <>
+      <div style={{
+        display: 'grid',
+        gap: 5,
+        gridTemplateColumns: '6fr 1fr',
+        marginBottom: 5
+      }}>
+        {typeof currentValue !== 'string' &&
+          currentValue.map((l) => (
```
binh-dam-ibigroup, line 48: This statement is failing on my machine with `currentValue` being `null` when creating a new location group in an existing editor db with a flex feed already loaded, or when loading a flex feed into a brand new editor instance.

lib/editor/components/LocationGroupsEditor.js
```diff
@@ -0,0 +1,92 @@
...
+        entityKey={'stop_or_location_id'}
+        entities={[
+          ...stops,
+          ...locations.map((l) =>
+            // id is needed here to avoid conflicts with stop ids
+            ({ ...l, id: uuidv4() })
+          )
+          // $FlowFixMe Flow struggles with union types
+        ].filter((stopOrLocation: GtfsStop | GtfsLocation) => {
+          return !currentValue.includes(
```
binh-dam-ibigroup: A null check is needed here too (I think it should be true if `currentValue` is `null`).

lib/editor/components/LocationGroupsEditor.js
```diff
@@ -0,0 +1,92 @@
...
+            </React.Fragment>
+          ))}
+      </div>
+      {!dropdownShowing && !(stops.length === 0 && locations.length === 0) && <Button
+        block
+        bsSize='small'
+        onClick={() => setDropdownShowing(true)}
+      >
+        <Icon type='plus' /> Add stop or location by name
+      </Button>}
```
binh-dam-ibigroup, line 63: Add some text above the button saying no stop/location have been added yet - Should that be a validation error?

lib/editor/components/LocationGroupsEditor.js
```diff
@@ -0,0 +1,92 @@
...
+          // $FlowFixMe Flow struggles with union types
+        ].filter((stopOrLocation: GtfsStop | GtfsLocation) => {
+          return !currentValue.includes(
+            // $flowfixme making this flow compatible would introduce a lot of unneeded code
+            stopOrLocation.stop_id || stopOrLocation.location_id
+          )
+        })}
+        onChange={(change) => {
+          processFieldChange([
+            ...currentValue,
```
binh-dam-ibigroup: A null check seems needed here too.

lib/editor/components/EditorInput.js
```diff
@@ -400,6 +399,22 @@ export default class EditorInput extends React.Component<Props> {
           </FormGroup>
         )
       }
+      case 'GTFS_STOP_OR_LOCATION_LIST': {
+        const stops: Array<GtfsStop> = getTableById(tableData, 'stop')
+        const locations: Array<GtfsLocation> = getTableById(tableData, 'location')
+
+        return (
+          <FormGroup {...formProps}>
+            {basicLabel}
```
binh-dam-ibigroup: I would forego the field name `location_id` which does not correspond to the elements being shown, and use "Locations" instead.

lib/editor/actions/active.js
```diff
@@ -413,6 +413,14 @@ export function saveEntity (
...
     const data = mappingStrategy(entityWithDefaults)
 
+    if (data.location_group_id && data.location_id) {
+      data.location_id = data.location_id.join(',')
+      if (data.is_creating) {
+        delete data.is_creating
+        delete data.id
+      }
+    }
+
```
binh-dam-ibigroup, line 423: Clicking on a location group in the `EntityList` does not always show the editor for the clicked location group. A console log mentions: "subComponent: 'undefined' is not a valid URL subComponent path" (see line 190), and as a result, the URL is not updated in those instances. The reason for that is that `entityId` is not really defined for a `LocationGroup`, or is it?

binh-dam-ibigroup: This causes a strange behavior where the target url is "committed" to the router but not in the browser.

binh-dam-ibigroup (commented):
I am kinda stuck with these strange errors, could you please advise?

lib/editor/components/LocationGroupsEditor.js
miles-grant-ibigroup: Looks like this didn't work https://github.com/ibi-group/datatools-ui/pull/812#discussion_r934903531. Have fixed with an explicit null check

lib/editor/components/LocationGroupsEditor.js
miles-grant-ibigroup: I think flex best practices validation errors are on our roadmap. An empty location group complies with the spec so it is ok for now

lib/editor/actions/active.js
miles-grant-ibigroup: I was able to reproduce this once, but then not again. I believe this is a more general datatools bug, as I can't see how location groups differ from any other entity. Do you have any ideas?

miles-grant-ibigroup: I've debugged this some more and it appears to be related to the data tools-server version used. Since a lot of the location group server PRs have yet to be merged, I would wait until this PR is officially unblocked before debugging this further.

miles-grant-ibigroup:
Thanks for the review Binh -- that white screen of death bug should be resolved now.

lib/editor/components/map/PatternLocationMarker.js
```diff
@@ -110,44 +115,51 @@ export default class PatternLocationMarker extends Component<Props> {
...
-                    <FormGroup
-                      controlId='defaultTimeInLocation'>
-                      <ControlLabel>Time in location</ControlLabel>
-                      <MinuteSecondInput
-                        seconds={patternLocation.flexDefaultZoneTime}
-                        onChange={this._onChangeTimeInLocation} />
-                    </FormGroup>
-                  </Col>
-                </Row>
+                { // If we do not have a patternLocation, do not render patternLocation edit buttons
```
binh-dam-ibigroup, line 132: What is the reason for not doing so? The action buttons and times are already available in the pattern editor for a location group.

lib/editor/components/timetable/TimetableEditor.js
```diff
@@ -18,7 +18,7 @@ import type {Feed, FetchStatus, Pattern, StopTime, TimetableColumn, Trip, TripCo
 import type {TripValidationIssues} from '../../selectors/timetable'
 import type {EditorTables, TimetableState} from '../../../types/reducers'
 import { mergePatternHaltsOfPattern } from '../../../gtfs/util'
-import { patternHaltIsStop } from '../../util/location'
+import { patternHaltIsLocation, patternHaltIsStop } from '../../util/location'
```
binh-dam-ibigroup, line 21: After adding a location group in a route pattern, the UI doesn't seem to pick up changes to the cells for the location group, so it is not possible to edit the scheduled times at location groups.

binh-dam-ibigroup (changes requested):
Two general comments, one on action buttons for location groups on the map popup, and one about editing schedules for location groups.

lib/editor/components/map/PatternLocationMarker.js
miles-grant-ibigroup: We did this thinking it might be confusing because there was no good way to make it clear when you were editing a location or a location group, as the popup would be the same and could potentially include the same location twice on the map.

lib/editor/components/timetable/TimetableEditor.js
miles-grant-ibigroup: I've tried to but failed to replicate this. Are you able to provide more detailed steps to reproduce?

miles-grant-ibigroup:
Unblocking this as the server end of this is merged into `dev-flex`.

miles-grant-ibigroup:
@binh-dam-ibigroup the bug you reported was reproducible, but with [the latest datatools-server](https://github.com/ibi-group/datatools-server/pull/480) things are working again.

Unblocking as the server-side of this is merged

lib/editor/actions/tripPattern.js
```diff
@@ -79,28 +79,31 @@ export function normalizeStopTimes (patternId: number, beginStopSequence: number
...
     const sortedPatternHalts = patternHalts.map(resequenceStops)
-    const {stops, locations} = sortedPatternHalts.reduce(
+    const {stops, locationGroups, locations} = sortedPatternHalts.reduce(
       (acc, cur) => {
         if (patternHaltIsLocation(cur)) {
           acc.locations.push(cur)
         }
+        if (patternHaltIsLocationGroup(cur)) {
+          acc.locationGroups.push(cur)
+        }
```
binh-dam-ibigroup, line 93: Either here or in `PatternStopCard`, when creating/saving a pattern stop that is a location or location group, the pickup type and drop-off type should be checked for the value zero which is invalid. Otherwise, when importing an editor snapshot, a validation error about pickup and drop-off types will occur.

binh-dam-ibigroup (changes requested):
Sorry, I found a missing check when saving a pattern stop that is a location/location group, that would result in a validation error when exporting/importing an editor snapshot. Hopefully that's the last big glitch for now 🤞.

lib/editor/actions/tripPattern.js
miles-grant-ibigroup: This is fixed in ca3070cbed9af18ff623b8ed253fff49bdbd748e, however only for new feeds. Existing feeds will still be exported incorrectly.

lib/editor/util/map.js
```diff
@@ -775,7 +775,7 @@ export function stopToPatternStop (
       continuousDropOff: 1,
       continuousPickup: 1,
       dropOffBookingRuleId: null,
-      dropOffType: 0,
+      dropOffType: 3,
```
binh-dam-ibigroup, line 778: Are the pickup/drop-off types fields meant to be editable in the pattern stop card (with a restricted set of allowed values)?

binh-dam-ibigroup (approved):
Looks like we have reached some kind of working state, so thanks for all the changes!

lib/editor/util/map.js
miles-grant-ibigroup: No. We decided that if you need to change this you should do so via booking rules, as you otherwise have two ways of changing the same thing.

We can always change this in the future but as there is no real-world use of this yet we thought doing this would set a good standard for newly created data.

lib/editor/components/EditorInput.js
```diff
@@ -400,6 +399,22 @@ export default class EditorInput extends React.Component<Props> {
           </FormGroup>
         )
       }
+      case 'GTFS_STOP_OR_LOCATION_LIST': {
```
daniel-heppner-ibigroup, line 402: This needs to be added to the `InputType` union in `types/index.js`.

daniel-heppner-ibigroup (approved):
It works, looks like Binh caught a lot of the bugs. Functionally works great for me. I found a type issue that can be fixed too.

lib/editor/components/EditorInput.js
miles-grant-ibigroup: Wow that's a really good catch thank you! Updated in 395acdfd83e4294b06f4cbb347c9e928ffab223e

PR #817, opened 2022-07-13
fix(TimetableHeader): separate offset input into two inputs

miles-grant-ibigroup:
### Checklist

- [X] The description lists any configuration setting(s) that differ from the default settings
- [X] All tests and CI builds passing

### Description

This PR refactors the offset input to be 2 separate textboxes. This removes the previous behavior in the unified textbox, which was seen by many to be confusing and complicated.

The input is now separated between hours and minutes.

Screen Shot 2022-07-13 at 3 51 50 PM

One piece of functionality I feel is missing is automatically moving to the minute input field, either after the user types a `:` or perhaps after they type two characters? Curious what others think about this.

### DIFF
--- a/lib/editor/components/HourMinuteInput.js
+++ b/lib/editor/components/HourMinuteInput.js
@@ -1,12 +1,13 @@
 // @flow
 
-import React, {Component} from 'react'
+// $FlowFixMe Flow doesn't recognize react KeyboardEvent
+import React, {Component, KeyboardEvent } from 'react'
 import {FormControl} from 'react-bootstrap'
 
 type Props = {
   onChange: (number) => any,
   seconds: number,
-  style: {[string]: string | number}
+  style?: {[string]: string | number}
 }
 
 type State = {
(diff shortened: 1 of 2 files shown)


miles-grant-ibigroup:
@philip-cline assigning this to you now, but since you're a more active datatools user than I am please let me know if the behavior is lacking/working unexpectedly and we can tweak this a little bit!

philip-cline:
I would definitely like to see the colon behaviour. I noted a few other things: 
- It would be nice to have "hh:mm" displayed as a placeholder when you first enter the timetable editor to indicate the input units. 
- I think some spacing between the inputs might look a bit better, with some corner rounding as well (perhaps on the offset button too). Something like: 
image

philip-cline (changes requested):
Couple things:
- We could remove the border radius on the rightmost input box so it flows into the offset button. 
- Typing a # over 60 in the minutes box will convert to hours, but keep the value in the minutes box. So typing 120 for example adds 2 hours, but ends up as 2:120. 
- The placeholder is showing "hh:mn" right now, and is cut off a bit.

miles-grant-ibigroup:
What browser are you using to get the `hh:mn` bug? Everything else I can replicate but this is how the textbox renders for me in Safari, Firefox, and Chrome
Screen Shot 2022-07-14 at 4 03 57 PM

philip-cline:
Ok that was my bad, I needed to reload the CSS files. I'll give another review

lib/editor/components/HourMinuteInput.js
```diff
@@ -1,48 +1,102 @@
 // @flow
 
-import React, {Component} from 'react'
+// $FlowFixMe Flow doesn't recognize react KeyboardEvent
+import React, {Component, KeyboardEvent } from 'react'
```
philip-cline:
```suggestion
import React, {Component, KeyboardEvent} from 'react'
```

lib/editor/components/HourMinuteInput.js
```diff
@@ -1,48 +1,102 @@
...
+        />
+        <span style={{zIndex: 9, position: 'relative', left: -2.5, fontWeight: 900}}>:</span>
+        <FormControl
+          {...otherProps}
+          id='minutes'
+          onKeyDown={this._onKeyDown}
+          onChange={this._onChange}
+          onFocus={this._onFocus}
+          placeholder={'mm'}
+          ref={'minuteInput'}
```
philip-cline: Is this `ref` used?

philip-cline (approved):
Everything looks good and works well. Just a few small comments

lib/editor/components/HourMinuteInput.js
miles-grant-ibigroup: It is not! Good catch thank you

miles-grant-ibigroup: Removed in 886c24350363c7266590e2d0e17650e30bd43dc7

daniel-heppner-ibigroup (approved):
This feels like the best solution to me!

Issue #549, opened 2020-04-02
Add date range selector for schedule exceptions in Editor


landonreed:
Currently when creating a schedule exception in the GTFS Editor, if the exception is intended to last for an extended period, each date must be added individually, which can be very tedious for any exception lasting over a week. We should replace the individual dates with a range selector (perhaps https://www.npmjs.com/package/react-date-range?).

![image](https://user-images.githubusercontent.com/2370911/78256616-c6731980-74c6-11ea-90bf-eecdaf0ed4b8.png)

binh-dam-ibigroup:
Closed by #828.

PR #828, opened 2022-07-26
Add exception ranges

philip-cline:
### Checklist

### Description

This PR aims to implement #549, adding a range selector to the schedule exception form so that a series of dates can be easily entered. Current behaviour includes some awkward merging or dropping of dates if they are included in an existing range, which could be jarring to a user. It is intended to alert the user to these behaviours with `toast` popups in a future PR


lib/common/util/exceptions.js
```diff
@@ -0,0 +1,130 @@
...
+  else return false
+}
+
+function previousDateIsConsecutive (prev: ?string, currentDate: moment): boolean {
+  if (!prev) return false
+  const prevDate = moment(prev)
+
+  if (prev && currentDate.diff(prevDate, 'days') === 1) return true
+  else return false
+}
```
miles-grant-ibigroup: This method is identical to `datesAreConsecutive` and can probably be removed

lib/common/util/exceptions.js
```diff
@@ -0,0 +1,130 @@
...
+      continue
+    }
+    datesToRemove.push(startRange.format('YYYYMMDD'))
+    startRange.add(1, 'days')
+  }
+  dates = (dates.filter(date => !datesToRemove.some(el => el === date)): Array<string>)
+  return dates
+}
+
+function datesAreConsecutive (next: ?string, currentDate: moment): boolean {
```
miles-grant-ibigroup: Based on how this is used, this method can probably use [rest parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters) as well as recursion to determine if a series of dates are consecutive.

lib/editor/components/ExceptionDate.js
```diff
@@ -46,31 +93,72 @@ export default class ExceptionDate extends Component<Props> {
     const {date, index, isNotValid} = this.props
     const dateTimeProps = {
       mode: 'date',
-      dateTime: date ? +moment(date) : +moment(),
+      dateTime: date ? +moment(date) : undefined,
```
miles-grant-ibigroup, line 95: Can probably be shortened to `date && +moment(date)`? Unless the library doesn't like `false`

lib/editor/components/ExceptionDate.js
```diff
@@ -46,31 +93,72 @@ export default class ExceptionDate extends Component<Props> {
...
+        </div>
+
+        <Button
+          disabled={!date}
+          onClick={this._addRange}
+          size='sm'
+          style={{
+            height: '100%',
+            textAlign: 'center',
+            gridColumn: '7 / span 5',
```
miles-grant-ibigroup: nitpick: you can probably write `12` instead of `span 5`

lib/editor/components/ExceptionDate.js
```diff
@@ -46,31 +93,72 @@ export default class ExceptionDate extends Component<Props> {
...
+            lineHeight: 0,
+            fontSize: 'x-small',
+            padding: '4px 4px',
```
miles-grant-ibigroup: can be shortened to `4`

lib/editor/components/ExceptionDateRange.js
```diff
@@ -0,0 +1,175 @@
...
+      (isStartDate && startDate === newRangeBoundary.format('YYYYMMDD')) ||
+      (!isStartDate && endDate === newRangeBoundary.format('YYYYMMDD'))
+    ) {
+      return
+    } else if (
+      (isStartDate && newRangeBoundary.diff(endMoment) > 0) ||
+      (!isStartDate && newRangeBoundary.diff(startMoment) < 0)
+    ) {
+      return
+    }
```
miles-grant-ibigroup: Might be cleaner to combine these into one large if block?

lib/editor/components/ExceptionDateRange.js
```diff
@@ -0,0 +1,175 @@
...
+
+    // Filter out new dates (0s), add them back in later to preserve new dates in final array
+    const newDates = activeEntity.dates.filter(date => typeof date === 'number')
+    // $FlowFixMe Flow does not recognize filtering on typeof
+    let dates = ((activeEntity.dates.filter(date => typeof date === 'string') || []): Array<string>)
```
miles-grant-ibigroup: Why does `dates` have the `|| []` hack but `newDates` doesn't?

lib/editor/components/ExceptionDateRange.js
```diff
@@ -0,0 +1,175 @@
...
+      dates = rangeExtended
+        ? addRangeOfDates(newRangeBoundary, startMoment, dates, true, false)
+        : removeRangeOfDates(startMoment, newRangeBoundary, dates, true, false)
+    } else {
+      const rangeExtended = newRangeBoundary.diff(endMoment) > 0
+      dates = rangeExtended
+        ? addRangeOfDates(endMoment, newRangeBoundary, dates, false, true)
+        : removeRangeOfDates(newRangeBoundary, endMoment, dates, false, true)
+    }
+
```
miles-grant-ibigroup: I would move the complexity of handling this to the utils file. This would also make the arg list cleaner

lib/editor/components/ExceptionDateRange.js
```diff
@@ -0,0 +1,175 @@
...
+    updateActiveGtfsEntity({
+      component: activeComponent,
+      entity: activeEntity,
+      props: {dates: [...dates, ...newDates]}
+    })
+  }
+
+  render () {
+    const {range, index, isNotValid} = this.props
```
miles-grant-ibigroup: nitpick: can we rename this to `isInvalid` (or flip its value to `isValid`)? Otherwise the double negation gets confusing

lib/editor/components/ScheduleExceptionForm.js
```diff
@@ -84,6 +87,28 @@ export default class ScheduleExceptionForm extends Component<Props> {
...
+    const newDates = dates.reduce((acc, cur, curIndex) => {
+      if (typeof cur === 'number') acc.push({date: 0, index: curIndex})
+      return acc
+    }, [])
+
+    dates = filterOutNewDates(dates)
+    dates.sort((a, b) => moment(a).diff(moment(b)))
+
+    // $FlowFixMe Flow does not understand proper typing for array#reduce
+    const {ranges, dates: parsedDates} = dates.reduce(createDateRangesFromDates, {ranges: [], dates: []})
```
miles-grant-ibigroup: I understand the desire to make things cleaner by breaking the process up into steps, but I think this would be cleaner as a single reducer -- the steps can be split up inside that reducer

lib/editor/components/ScheduleExceptionForm.js
```diff
@@ -100,6 +125,12 @@ export default class ScheduleExceptionForm extends Component<Props> {
     ? 'error'
     : undefined
 
+  _checkRangeValidation = (startIndex: number, endIndex: number) => {
+    // Rather than loop through indices, create an array of indices to consolide checks
```
miles-grant-ibigroup: Typo?
```suggestion
    // Rather than loop through indices, create an array of indices to consolidate checks
```

lib/editor/components/ScheduleExceptionForm.js
```diff
@@ -100,6 +125,12 @@ export default class ScheduleExceptionForm extends Component<Props> {
...
+    const rangeIndices = Array.from({length: endIndex - startIndex + 1}, (el, index) => startIndex + index)
+    return rangeIndices.some(index => Boolean(this._checkValidation(`dates-${index}`)))
```
miles-grant-ibigroup: I see what you're trying to do here, but this is just as many passes over the array as just doing a for loop! I do love functional programming though so this has my blessing

lib/editor/components/ScheduleExceptionForm.js
```diff
@@ -197,18 +231,26 @@ export default class ScheduleExceptionForm extends Component<Props> {
...
-                  index={index}
-                  key={index}
-                  date={date}
-                  // Exception dates expects a boolean
-                  isNotValid={Boolean(this._checkValidation(`dates-${index}`))}
-                  {...this.props}
-                />
-              ))
-              : <div>No dates specified</div>
+            {ranges && ranges.length !== 0 && ranges.map((range, index) => (
```
miles-grant-ibigroup: You shouldn't need this length check -- mapping an empty array should return nothing

lib/editor/components/ScheduleExceptionForm.js
```diff
@@ -197,18 +231,26 @@ export default class ScheduleExceptionForm extends Component<Props> {
...
+              <ExceptionDateRange
+                index={index}
+                key={index}
+                range={range}
+                isNotValid={this._checkRangeValidation(range.startIndex, range.endIndex)}
+                {...this.props}
+              />
+            ))}
+            {parsedDates && parsedDates.length !== 0 && parsedDates.map((date, index) => (
```
miles-grant-ibigroup: Same as above with the length check

lib/types/index.js
```diff
@@ -826,7 +826,7 @@ export type ReactSelectOptions = Array<ReactSelectOption>
 export type ScheduleException = {|
   added_service: ?Array<string>,
   custom_schedule: ?Array<string>,
-  dates: Array<string>,
+  dates: Array<string | 0>,
```
miles-grant-ibigroup: You should add a comment here about why this is happening and also perhaps extract `string | 0` to a new type

lib/common/util/exceptions.js
```diff
@@ -0,0 +1,130 @@
...
+// Method to filter out new dates (0's) so that flow is happy
+// We have to use a reducer to keep flow happy.
+export function filterOutNewDates (dates: Array<string | 0>): Array<string> {
+  const outputDates: Array<string> = []
+  const filter = (acc, current) => {
+    typeof current === 'string' && acc.push(current)
+    return acc
+  }
+  dates.reduce(filter, outputDates)
+  return outputDates
```
miles-grant-ibigroup: Couldn't this entire method just be one `filter` call?

miles-grant-ibigroup: Am I missing something or would this work? 
```js
return dates.filter(date => typeof date === 'string')
```

lib/common/util/exceptions.js
```diff
@@ -0,0 +1,130 @@
...
+}
+
+/**
+ * This method adds a range of dates from startRange to endRange to the dates array, inclusively according to startInclusive
+ * and endInclusive.
+ */
+export function addRangeOfDates (
+  startRange: moment,
```
miles-grant-ibigroup: Is `moment` a valid flow type?

lib/common/util/exceptions.js
```diff
@@ -0,0 +1,130 @@
...
+  }
+  dates.reduce(filter, outputDates)
+  return outputDates
+}
+
+/**
+ * This method adds a range of dates from startRange to endRange to the dates array, inclusively according to startInclusive
+ * and endInclusive.
+ */
+export function addRangeOfDates (
```
miles-grant-ibigroup: I would refactor this method to be a recursive method which only accepts `dates` and `endRange`. The method would return `dates` if `endRange` is in `dates`.  I'd then also create a second method to do the same but with `startRange`. These methods can then be combined to re-create the current functionality with less complexity

lib/common/util/exceptions.js
```diff
@@ -0,0 +1,130 @@
...
+    }
+  })
+  return dates
+}
+
+/**
+ * This method removes a range of dates from startRange to endRange to the dates array, inclusively according to startInclusive
+ * and endInclusive.
+ */
+export function removeRangeOfDates (
```
miles-grant-ibigroup: Like `addRangeOfDates`, I would make this a recursive method which should reduce complexity and the number of arguments needed. Since we have a pair of methods now, we could even add a flag to those methods to switch them between add and remove modes

lib/common/util/exceptions.js
```diff
@@ -0,0 +1,130 @@
...
+
+  if (prev && currentDate.diff(prevDate, 'days') === 1) return true
+  else return false
+}
+
+/**
+ * This method extracts any ranges from a list of dates that are either being generated
+ * or obtained from the back end.
+ */
+export function createDateRangesFromDates (acc: ParsedRangesAndDates, current: string, currentIndex: number, dates: Array<string>) {
```
miles-grant-ibigroup: If we're going to have a method this complicated, then we need tests for it. 

I personally would be in favor of again using recursion to clean this up. With one list of range starts, one list of range ends, and a list of dates, the dates array can be processed without needing `current` or `currentIndex`.

miles-grant-ibigroup (changes requested):
Functionality is good, but there are some bugs that will probably be removed by cleaning up the methods in the utils folder.

My comments about the react are mostly nitpicks, no major issues there. Some issues of complexity creeping into the React, but that will be resolved by refactoring the utils methods.

For the utils functions there are a lot of possible refactorings to be done. After having thought about it for a bit, I've convinced myself that recursion is the best way to clean them up, however I'm open to a more reducer-based refactoring as well. There are a lot of methods that right now rely on keeping track of a series of indices across multiple loopings of various arrays. I think moving to recursion can clean up these methods and force us to write cleaner code. Majorly refactoring those util methods should also fix the most noticeable bugs as well.

Some of my refactoring suggestions are quite hand-wavy so if you'd like me to make a piggyback PR/discuss refactoring ideas outside of github/etc please let me know!

lib/editor/components/ExceptionDateRange.js
philip-cline: Hm yeah flow was being confusing. I removed the hack from the second one since it's being supressed anyway.

lib/editor/components/ExceptionDateRange.js
philip-cline: Sure, will be updated in the next commit

lib/editor/components/ExceptionDate.js
philip-cline: 👍

lib/common/util/exceptions.js
philip-cline: I don't think rest parameters will be necessary, I think this can be done just with an absolute diff

lib/editor/components/ExceptionDate.js
philip-cline: Yeah unfortunately the library rly doesn't like `false`, it interprets it as a 0, or Jan 1 1970

lib/editor/components/ScheduleExceptionForm.js
philip-cline: Yeah for sure, I just found it a little more concise and readable (maybe).

lib/common/util/exceptions.js
philip-cline: This was Flow confusion, but we're just giving up now and using `$FlowFixMe`s now since Flow doesn't understand basic array methods

lib/common/util/exceptions.js
philip-cline: Yep, I was basing this off of [this](https://stackoverflow.com/questions/43036634/how-to-specify-a-moment-js-object-annotation-in-flow).

lib/common/util/exceptions.js
philip-cline: `datesAreConsecutive` is now fully removed w/ the latest recursive methods in https://github.com/ibi-group/datatools-ui/pull/828/commits/0bd970887b85994ed9f9452ec5640b01eb021f28!

lib/common/util/exceptions.js
philip-cline: `datesAreConsecutive` is now fully removed w/ the latest recursive methods in https://github.com/ibi-group/datatools-ui/pull/828/commits/0bd970887b85994ed9f9452ec5640b01eb021f28!

lib/editor/components/ExceptionDate.js
philip-cline: :) done in https://github.com/ibi-group/datatools-ui/pull/828/commits/0bd970887b85994ed9f9452ec5640b01eb021f28

lib/editor/components/ExceptionDate.js
philip-cline: Fixed in https://github.com/ibi-group/datatools-ui/pull/828/commits/0bd970887b85994ed9f9452ec5640b01eb021f28

lib/editor/components/ExceptionDateRange.js
philip-cline: One block in https://github.com/ibi-group/datatools-ui/pull/828/commits/0bd970887b85994ed9f9452ec5640b01eb021f28

lib/editor/components/ExceptionDateRange.js
philip-cline: Removed in https://github.com/ibi-group/datatools-ui/pull/828/commits/0bd970887b85994ed9f9452ec5640b01eb021f28

lib/editor/components/ExceptionDateRange.js
philip-cline: Good call, this is in its own method `updateDates` in the utils file

lib/editor/components/ExceptionDateRange.js
philip-cline: Yep, renamed in https://github.com/ibi-group/datatools-ui/pull/828/commits/0bd970887b85994ed9f9452ec5640b01eb021f28

lib/editor/components/ScheduleExceptionForm.js
philip-cline: Sorting was necessary before using the reducer, but this whole method has been cleaned up into a recursive method in https://github.com/ibi-group/datatools-ui/pull/828/commits/0bd970887b85994ed9f9452ec5640b01eb021f28

lib/editor/components/ScheduleExceptionForm.js
philip-cline: Typo. fixed in https://github.com/ibi-group/datatools-ui/pull/828/commits/0bd970887b85994ed9f9452ec5640b01eb021f28

lib/editor/components/ScheduleExceptionForm.js
philip-cline: Didn't change this in https://github.com/ibi-group/datatools-ui/pull/828/commits/0bd970887b85994ed9f9452ec5640b01eb021f28

lib/editor/components/ScheduleExceptionForm.js
philip-cline: Good call, removed in https://github.com/ibi-group/datatools-ui/pull/828/commits/0bd970887b85994ed9f9452ec5640b01eb021f28

lib/editor/components/ScheduleExceptionForm.js
philip-cline: Also removed in https://github.com/ibi-group/datatools-ui/pull/828/commits/0bd970887b85994ed9f9452ec5640b01eb021f28

lib/types/index.js
philip-cline: Sure, good call. I extracted to `ExceptionDate` in https://github.com/ibi-group/datatools-ui/pull/828/commits/0bd970887b85994ed9f9452ec5640b01eb021f2

lib/editor/components/ScheduleExceptionForm.js
miles-grant-ibigroup: excellent work

lib/common/util/exceptions.js
philip-cline: Changed in https://github.com/ibi-group/datatools-ui/pull/828/commits/0bd970887b85994ed9f9452ec5640b01eb021f28

lib/common/util/exceptions.js
philip-cline: You bet this is recursive now in https://github.com/ibi-group/datatools-ui/pull/828/commits/0bd970887b85994ed9f9452ec5640b01eb021f28

lib/common/util/exceptions.js
philip-cline: combined these methods and made recursive in https://github.com/ibi-group/datatools-ui/pull/828/commits/0bd970887b85994ed9f9452ec5640b01eb021f28

lib/common/util/exceptions.js
philip-cline: This method was cleaned up a lot in https://github.com/ibi-group/datatools-ui/pull/828/commits/0bd970887b85994ed9f9452ec5640b01eb021f28!

philip-cline:
Jumped the gun a bit, I'm gonna update the e2e tests for the new exception ranges

lib/common/util/exceptions.js
```diff
@@ -0,0 +1,69 @@
...
+
+import moment from 'moment'
+
+import type {ExceptionDate, ScheduleExceptionDateRange} from '../../types'
+
+/**
+ * This method updates the current dates based on the updated range. The method determines if the range has been extended or
+ * condensed and adds or subtracts dates as appropriate.
+ */
+export function updateDates (isStartDate: boolean, newRangeBoundary: moment, dates: Array<string>, startMoment: moment, endMoment: moment) {
```
miles-grant-ibigroup, line 11: I'm still not super happy with this method -- it seems to do the same thing just with different variables each time. Would it be possible to remove `isStartDate` and then replicate the functionality by passing in different variables?

philip-cline:
E2E tests will fail w/ the project deletion bug

lib/common/util/exceptions.js
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * This method parses dates to extract ranges and individual exception dates. It assumes that dates is sorted.
+ */
+export function getRangesForDates (dates: Array<ExceptionDate>, ranges: Array<ScheduleExceptionDateRange> = []) {
+  // If dates is empty, return ranges
+  if (dates.length === 0) return ranges
+  // Grab first date, iterate over ranges and see if any startRange is above the date
+  const date = dates.shift()
+  if (date === 0) ranges.push({startDate: 0, endDate: 0})
```
miles-grant-ibigroup, line 61: This method is well commented so we should probably add one more comment here that says why we did this

lib/editor/components/ExceptionDateRange.js
```diff
@@ -0,0 +1,162 @@
...
+    dates = (dates.filter(date => typeof date === 'string'): Array<string>)
+
+    const startMoment = moment(range.startDate)
+    const endMoment = moment(range.endDate)
+
+    if (key === 'range') {
+      dates = modifyRangeOfDates(dates, endMoment, startMoment, true)
+    } else if (key === 'end-date') {
+      dates = modifyRangeOfDates(dates, endMoment, startMoment.add(1, 'days'), true)
+    }
```
miles-grant-ibigroup: This might be cleaner as one function call that uses variables that are set based on the key

lib/editor/components/ScheduleExceptionForm.js
```diff
@@ -197,18 +216,26 @@ export default class ScheduleExceptionForm extends Component<Props> {
...
+                isInvalid={this._checkRangeValidation(dates.findIndex(date => date === range.startDate), dates.findIndex(date => date === range.endDate))}
+                {...this.props}
+              />
+            ))}
+            {parsedDates && parsedDates.map((date, index) => (
+              <ExceptionDate
+                index={dates.findIndex(d => d === date)}
+                key={index}
+                date={date}
+                isInvalid={Boolean(this._checkValidation(`dates-${dates.findIndex(d => d === date)}`))}
```
miles-grant-ibigroup: We can extract `dates.findIndex(d => d === date)`

miles-grant-ibigroup (changes requested):
Just a few small changes, mostly nitpicks!

lib/common/util/exceptions.js
philip-cline: refactoring w/o `isStartDate` would look something like this: 
image
image
which imo is not much cleaner or easier to understand. Miles and I discussed a bit and we'll leave it to the second reviewer to break the tie on this

miles-grant-ibigroup (approved):
Looking forward to the second review on this

lib/common/util/exceptions.js
philip-cline: updated in https://github.com/ibi-group/datatools-ui/pull/828/commits/b8ea974dc8ba296b92d2b964760b893be231af36

lib/editor/components/ExceptionDateRange.js
philip-cline: good call, this is much cleaner in https://github.com/ibi-group/datatools-ui/pull/828/commits/b8ea974dc8ba296b92d2b964760b893be231af36

lib/editor/components/ScheduleExceptionForm.js
philip-cline: updated in https://github.com/ibi-group/datatools-ui/pull/828/commits/b8ea974dc8ba296b92d2b964760b893be231af36

lib/editor/components/ExceptionDate.js
```diff
@@ -5,25 +5,73 @@ import React, {Component} from 'react'
...
+
+    // Check if the date selected is already in an existing range.
+    // Invalid date results from typing in the Datetimepicker component (which is valid as long as it resolves to a date)
+    if (dates.some(date => date === newDate && date !== 'Invalid date')) {
+      dates.splice(index, 1)
+      toast.warn(
+        `ⓘ Date has been removed. Date entered is already included in an existing range or single date!`,
+        {
+          position: 'top-right',
+          autoClose: 4000,
```
binh-dam-ibigroup, line 63: Sort style props

lib/editor/components/ExceptionDate.js
```diff
@@ -5,25 +5,73 @@ import React, {Component} from 'react'
...
 }
 
-export default class ExceptionDate extends Component<Props> {
+type State = {
+    isDeleted: boolean,
+    isDuplicate: boolean
+}
+
+export const inputStyleProps = {
+  fontSize: '10px',
```
binh-dam-ibigroup: At first sight I find using a smaller font difficult on the eyes...

lib/editor/components/ExceptionDate.js
```diff
@@ -43,34 +91,76 @@ export default class ExceptionDate extends Component<Props> {
...
         key={index}
-        style={{position: 'relative', width: '100%', marginBottom: '5px'}}>
+        style={{
+          alignItems: 'center',
+          display: 'grid',
+          gap: 5,
+          gridTemplateColumns: 'repeat(12, 1fr)',
+          justifyContent: 'center',
+          textAlign: 'center',
+          position: 'relative',
```
binh-dam-ibigroup, line 115: Sort style props

lib/editor/components/ExceptionDate.js
```diff
@@ -43,34 +91,76 @@ export default class ExceptionDate extends Component<Props> {
...
+          marginBottom: '5px'
+        }}>
+        <div
+          style={{gridColumn: '1 / 6'}}
+        >
```
binh-dam-ibigroup: Place this in one line.

lib/editor/components/ScheduleExceptionForm.js
```diff
@@ -197,18 +216,29 @@ export default class ScheduleExceptionForm extends Component<Props> {
...
-              ))
-              : <div>No dates specified</div>
+            {ranges && ranges.map((range, index) => {
+              const startDateIndex = dates.findIndex(d => d === range.startDate)
+              const endDateIndex = dates.findIndex(d => d === range.endDate)
+              return <ExceptionDateRange
+                index={index}
+                key={index}
+                range={range}
+                isInvalid={this._checkRangeValidation(startDateIndex, endDateIndex)}
```
binh-dam-ibigroup: Sort props (2 instances)

lib/editor/components/ScheduleExceptionForm.js
```diff
@@ -197,18 +216,29 @@ export default class ScheduleExceptionForm extends Component<Props> {
...
-                  // Exception dates expects a boolean
-                  isNotValid={Boolean(this._checkValidation(`dates-${index}`))}
-                  {...this.props}
-                />
-              ))
-              : <div>No dates specified</div>
+            {ranges && ranges.map((range, index) => {
+              const startDateIndex = dates.findIndex(d => d === range.startDate)
+              const endDateIndex = dates.findIndex(d => d === range.endDate)
+              return <ExceptionDateRange
```
binh-dam-ibigroup: I would surround the tag with parentheses (2 instances).

lib/types/index.js
```diff
@@ -822,11 +822,12 @@ export type ReactSelectOption = {
 }
 
 export type ReactSelectOptions = Array<ReactSelectOption>
+export type ExceptionDate = string | 0 // 0's are include in the date type for new dates.
```
binh-dam-ibigroup: Typo "included"

lib/editor/components/ExceptionDate.js
```diff
@@ -43,34 +91,76 @@ export default class ExceptionDate extends Component<Props> {
...
+      dateTime: date ? +moment(date) : undefined,
       onChange: this._onDateChange,
-      defaultText: undefined
+      defaultText: undefined,
+      size: 'sm'
     }
+
     if (!date) {
-      dateTimeProps.defaultText = 'Please select a date'
+      dateTimeProps.defaultText = 'Select date'
```
binh-dam-ibigroup, line 102: (Nitpick)Although this was already set as text before, IMO this should be the `placeholder` prop.

lib/editor/components/ExceptionDate.js
```diff
@@ -43,34 +91,76 @@ export default class ExceptionDate extends Component<Props> {
...
+
+        <Button
+          data-test-id='exception-add-range'
+          disabled={!date}
+          onClick={this._addRange}
+          size='sm'
+          style={{
+            height: '100%',
+            textAlign: 'center',
+            gridColumn: '7 / 12',
```
binh-dam-ibigroup, line 134: Sort style props.

lib/editor/components/ExceptionDate.js
```diff
@@ -43,34 +91,76 @@ export default class ExceptionDate extends Component<Props> {
...
+            lineHeight: 0,
+            fontSize: 'x-small',
```
binh-dam-ibigroup: For this button, there is probably no need to override the font size as there seems to be ample space to use the ambient font size.

lib/editor/components/ExceptionDate.js
```diff
@@ -43,34 +91,76 @@ export default class ExceptionDate extends Component<Props> {
...
+        </div>
+
+        <Button
+          data-test-id='exception-add-range'
+          disabled={!date}
+          onClick={this._addRange}
+          size='sm'
+          style={{
+            height: '100%',
+            textAlign: 'center',
```
binh-dam-ibigroup: This prop is already the default in most browsers.

lib/editor/components/ExceptionDate.js
```diff
@@ -43,34 +91,76 @@ export default class ExceptionDate extends Component<Props> {
...
+            gridColumn: '7 / 12',
+            lineHeight: 0,
+            fontSize: 'x-small',
+            padding: '4',
+            margin: 0,
+            width: '100%'
```
binh-dam-ibigroup: Is it necessary to override margin (zero by default) and width (set by grid columns, looks like)?

lib/editor/components/ExceptionDate.js
```diff
@@ -43,34 +91,76 @@ export default class ExceptionDate extends Component<Props> {
...
+          justifyContent: 'center',
+          textAlign: 'center',
+          position: 'relative',
+          marginBottom: '5px'
+        }}>
+        <div
+          style={{gridColumn: '1 / 6'}}
+        >
+          <DateTimeField
+            inputProps={{style: {...inputStyleProps, backgroundColor: isInvalid ? '#FBDEDE' : undefined}}}
```
binh-dam-ibigroup: Try using the react-bootstrap `validationState` prop instead of changing the background?

lib/editor/components/ExceptionDateRange.js
```diff
@@ -0,0 +1,159 @@
...
+      defaultText: undefined,
+      mode: 'date',
+      size: 'sm',
+      style: {
+        gridColumn: 'span 5'
+      }
+    }
+
+    // Highlight both inputs red if we have invalid input.
+    const rangeInputStyleProps = {...inputStyleProps, backgroundColor: isInvalid ? '#FBDEDE' : undefined}
```
binh-dam-ibigroup: Try using the react-bootstrap `validationState` prop instead of changing the background.

lib/editor/components/ExceptionDateRange.js
```diff
@@ -0,0 +1,159 @@
...
+    const startMoment = moment(startDate)
+    const endMoment = moment(endDate)
+
+    // This prevents the user from inputting invalid data. Validation is not required since this behaviour
+    // is fully prevented (see also use of minDate and maxDate in DateTimePickers)
+    if (
+      (isStartDate && (startDate === newRangeBoundary.format('YYYYMMDD') || (isStartDate && newRangeBoundary.diff(endMoment) > 0))) ||
+      (!isStartDate && (endDate === newRangeBoundary.format('YYYYMMDD') || newRangeBoundary.diff(startMoment) < 0))
+    ) {
+      return
```
binh-dam-ibigroup, line 51: FYI (non-blocking, but this can become an annoyance)This is resetting to the initial or last saved value (only in the start date) if the user (in)voluntarily enters an incorrect date.

Screen Recording 2022-08-15 at 3 21 57 PM

lib/common/util/exceptions.js
binh-dam-ibigroup: Entering an end date that is before the start date (or a start date that is after the end date) causes a stack overflow exception, which is probably preventing validation to occur (to allow saving or to show an error).

lib/editor/components/ExceptionDate.js
```diff
@@ -43,34 +91,76 @@ export default class ExceptionDate extends Component<Props> {
...
     }
     return (
       <div
         key={index}
-        style={{position: 'relative', width: '100%', marginBottom: '5px'}}>
+        style={{
+          alignItems: 'center',
+          display: 'grid',
+          gap: 5,
+          gridTemplateColumns: 'repeat(12, 1fr)',
```
binh-dam-ibigroup, line 112: Consider layout tweaks from #839.

lib/common/util/exceptions.js
binh-dam-ibigroup: Other than the problem above, the method is fine as written. No need to add more `?` to it.

lib/editor/components/ExceptionDate.js
binh-dam-ibigroup: See #839 for using the ambient font.

binh-dam-ibigroup (changes requested):
Good overall, with a few tweaks needed:
- please use the ambient font size (see #839)
- please check that after a chasnge, the start date of a range is not after the end date
- try to use the `validationState` for `inputProps` to render validation errors

lib/editor/components/ExceptionDate.js
philip-cline: Thanks for the suggestion, I merged #839 which helps a lot with the small fonts

lib/editor/components/ExceptionDate.js
philip-cline: It's passed to the `DateTimeField` which uses it as a placeholder basically, isn't this the same thing as using a placeholder prop, just through the DateTimeField API? 

the API is here: https://github.com/quri/react-bootstrap-datetimepicker

philip-cline: Maybe I'm confused about what you mean though, would you prefer passing directly in the props rather than in this props object?

binh-dam-ibigroup: `defaultText` will make the text editable, so portions of the "Select date" could remain after the user starts typing in the component.

philip-cline: Right, thanks I will mess around with that a bit.

lib/editor/components/ExceptionDateRange.js
philip-cline: Yeah thanks for pointing this out, there's also some other weird behaviour that you can trigger where the exception date is automatically populated, but not respected by the validation. I'm going to address this and those issues in a separate forthcoming PR because I suspect the solution will be similar, and the previous issue existed even in prod

lib/editor/components/ExceptionDate.js
philip-cline: Switching to a placeholder prop causes the datetimepicker to snap to the current date as a default, so it seems this may be necessary.

lib/common/util/exceptions.js
philip-cline: Sounds good, I will leave as written and stack overflow exception fixed in https://github.com/ibi-group/datatools-ui/pull/828/commits/7697c17f4dc28e0c1f53a67d55602974fec138d8

lib/editor/components/ExceptionDate.js
philip-cline: fixed in https://github.com/ibi-group/datatools-ui/pull/828/commits/7697c17f4dc28e0c1f53a67d55602974fec138d8

lib/editor/components/ExceptionDate.js
philip-cline: Fixed in https://github.com/ibi-group/datatools-ui/pull/828/commits/7697c17f4dc28e0c1f53a67d55602974fec138d8

lib/editor/components/ExceptionDate.js
philip-cline: Fixed in https://github.com/ibi-group/datatools-ui/pull/828/commits/7697c17f4dc28e0c1f53a67d55602974fec138d8

lib/editor/components/ScheduleExceptionForm.js
philip-cline: Fixed in https://github.com/ibi-group/datatools-ui/pull/828/commits/7697c17f4dc28e0c1f53a67d55602974fec138d8

lib/editor/components/ScheduleExceptionForm.js
philip-cline: Fixed in https://github.com/ibi-group/datatools-ui/pull/828/commits/7697c17f4dc28e0c1f53a67d55602974fec138d8

lib/types/index.js
philip-cline: Fixed in https://github.com/ibi-group/datatools-ui/pull/828/commits/7697c17f4dc28e0c1f53a67d55602974fec138d8

lib/editor/components/ExceptionDate.js
philip-cline: This is again related to an issue that's currently in prod, which I can tackle a little bit more in that PR

lib/editor/components/ExceptionDate.js
philip-cline: fixed in https://github.com/ibi-group/datatools-ui/pull/828/commits/7697c17f4dc28e0c1f53a67d55602974fec138d8

lib/editor/components/ExceptionDate.js
philip-cline: fixed in https://github.com/ibi-group/datatools-ui/pull/828/commits/7697c17f4dc28e0c1f53a67d55602974fec138d8

lib/editor/components/ExceptionDate.js
philip-cline: Fixed in https://github.com/ibi-group/datatools-ui/pull/828/commits/7697c17f4dc28e0c1f53a67d55602974fec138d8

lib/editor/components/ExceptionDate.js
philip-cline: Moved to bootstrap validation state in https://github.com/ibi-group/datatools-ui/pull/828/commits/7697c17f4dc28e0c1f53a67d55602974fec138d8 and added success state as well

lib/editor/components/ExceptionDateRange.js
philip-cline: Same as previous

lib/editor/components/ExceptionDate.js
philip-cline: Merged in #839

lib/editor/components/ExceptionDate.js
```diff
@@ -104,68 +102,64 @@ export default class ExceptionDate extends Component<Props, State> {
...
+            bsStyle='danger'
+            className='pull-right'
+            style={{
+              alignSelf: 'center',
+              gridColumn: '12',
+              height: '100%',
+              justifySelf: 'center',
+              padding: '2px 7.5px'
+            }}
+            key={`date-remove-${index}`}
```
binh-dam-ibigroup: Sort props.

lib/editor/components/ExceptionDate.js
```diff
@@ -104,68 +102,64 @@ export default class ExceptionDate extends Component<Props, State> {
...
-          display: 'grid',
-          gap: 5,
-          gridTemplateColumns: 'repeat(12, 1fr)',
-          justifyContent: 'center',
-          textAlign: 'center',
-          position: 'relative',
-          marginBottom: '5px'
-        }}>
+      <FormGroup validationState={validationState}>
         <div
```
binh-dam-ibigroup: Do we need this `<div>` now that `<FormGroup>` wraps everything? (We still need the props passed to that `<div>`)

lib/editor/components/ExceptionDateRange.js
```diff
@@ -15,14 +15,23 @@ type Props = {
...
-  isInvalid: boolean,
   range: ScheduleExceptionDateRange,
-  updateActiveGtfsEntity: typeof updateActiveGtfsEntity
+  updateActiveGtfsEntity: typeof updateActiveGtfsEntity,
+  validationState: string
 }
 
-export default class ExceptionDateRange extends Component<Props> {
+type State = {
+  validationStateOverride: boolean
```
binh-dam-ibigroup: Is this for forcing an error per the `render` method? Maybe rename to `forceErrorState`

lib/editor/components/ExceptionDateRange.js
```diff
@@ -76,84 +91,82 @@ export default class ExceptionDateRange extends Component<Props> {
...
-        style={{
-          display: 'grid',
-          gridGap: '5px',
-          gridTemplateColumns: 'repeat(12, 1fr)',
-          marginBottom: '5px'
-        }}
-      >
+      // Validation state override allows for control of input before date
+      <FormGroup validationState={validationState === 'error' || this.state.validationStateOverride ? 'error' : 'success'}>
         <div
```
binh-dam-ibigroup, line 116: Same as other comment, we could merge this `<div>` with the `<FormGroup>` above.

binh-dam-ibigroup (approved):
Much better now, good job!

lib/editor/components/ExceptionDate.js
philip-cline: sorted in https://github.com/ibi-group/datatools-ui/pull/828/commits/bd59f94bd41bd062d40630c054cacfdc97f32a98

lib/editor/components/ExceptionDate.js
philip-cline: Nope, good call. Removed and applied formatting to `<FormGroup>` in https://github.com/ibi-group/datatools-ui/pull/828/commits/bd59f94bd41bd062d40630c054cacfdc97f32a98

lib/editor/components/ExceptionDateRange.js
philip-cline: Yep, I like the rename. It's updated as of https://github.com/ibi-group/datatools-ui/pull/828/commits/bd59f94bd41bd062d40630c054cacfdc97f32a98

lib/editor/components/ExceptionDateRange.js
philip-cline: Same as above

PR #834, opened 2022-08-09
Auth0 react

### LINKED ISSUE #335: Authentication broken for safari
<a href="https://github.com/landonreed"></img></a> **Issue by [landonreed](https://github.com/landonreed)**
_Tuesday Jul 24, 2018 at 14:38 GMT_
_Originally opened as https://github.com/catalogueglobal/datatools-ui/issues/226_

----

It appears that [cross-origin authentication](https://auth0.com/docs/cross-origin-authentication) is broken for safari. Steps to recreate:

1. Open localhost in safari
2. Attempt to log in
3. `authorize` request to Auth0 fails (page will not load).

It looks like this authorize endpoint is also failing in other browsers (e.g. Chrome), but it does not have the effect of locking the user out.

### LINKED ISSUE #518: fetchAppInfo called every time a component mounts

Each time a component mounts, the app info (which contains config info and version info) is fetched:
https://github.com/ibi-group/datatools-ui/blob/af931edc5e9ee3cff446cb3b5d751853251a9390/lib/common/containers/App.js#L44-L56

## Expected behavior

This probably only needs to occur once -- when the first component mounts. The app info does not change (unless the server were to restart with a new config), so it likely does not need to be refetched each time a user navigates to a new route.

N/A

N/A

latest dev

binh-dam-ibigroup:
### Checklist

- [x] Appropriate branch selected => Using `react-17` because it has react 16+ required by `@auth0/auth0-react`.

### Description

This PR update the auth0 authentication library and refactors code accordingly.
Fix #518, Fix #335.

New front-end configuration parameters:
- `AUTH0_CONNECTION_NAME`: the name of the Auth0 connection used for logging in.

Required backend configuration parameters:
- `AUTH0_PUBLIC_KEY` must now be provided (`AUTH0_SECRET` will no longer be used).

Required changes in the Auth0 tenant:
- The algorithm for encoding the token must be set to `RS256` (this is required by the @auth0/auth0-react library).


lib/public/components/UserAccount.js
```diff
@@ -64,7 +64,12 @@ export default class UserAccount extends Component<Props> {
     if (user.profile) unsubscribeAll(user.profile)
   }
 
-  _onResetPassword = () => auth0.resetPassword()
+  _onResetPassword = () => {
+    if (window.confirm('This action will send you an email to change your password')) {
```
miles-grant-ibigroup: I don't think this confirm is needed. If it was an accident the user can just ignore the email, right?

lib/manager/actions/user.js
```diff
@@ -102,27 +84,12 @@ export function getRecentActivity (user: ManagerUserState) {
...
- *   an infinite loop
+ * Log user out from auth0 and from redux store.
  * @return {Action}  The userLogout action
  */
-export function logout (logoutFromAuth0: ?boolean = true) {
-  if (logoutFromAuth0) {
-    auth0.logout()
+export function logout (auth0Interface: Auth0ContextInterface) {
+  if (auth0Interface) {
+    auth0Interface.logout({ returnTo: window.location.origin })
```
miles-grant-ibigroup, line 92: This line seems to be causing issues in Firefox/Safari/Chrome. I'm getting a refresh loop even after clearing local storage/cookies. Every time the page loads I'm redirected to the auth0 logout page which then redirects back to localhost, etc.

It might be a configuration issue, but we should still avoid this behavior. Perhaps a cookie or local storage variable to signify that a redirect has already happened? A `noredirect` flag or similar on the target URL?

lib/common/containers/ActiveUserRetriever.js
```diff
@@ -0,0 +1,41 @@
...
+      // TRICKY: What we are passing to the backend is the id_token portion of the response.
+      // Unlike the access_token portion, the id_token contains the app_metadata info for the user.
+      receiveTokenAndProfile({ profile, token: tokenResponse.id_token })
+    } else if (!isLoading) {
+      logout(auth0)
+    }
+  }, [isAuthenticated, isLoading, profile])
+
+  // Component renders nothing.
+  return null
```
miles-grant-ibigroup, line 45: Given how this app is structured I don't entirely mind these "empty" components, but should we maybe combine them into a single one? Not sure how I feel about these empty components floating around. But it is quite clean as it is, so perhaps I'll leave this one up to the second reviewer

miles-grant-ibigroup (changes requested):
Will take a closer look once I get this working for me -- for now just a few high level questions

I'm hoping that the refresh loop is my fault not configuring things properly! Looking forward to seeing this in action

lib/common/containers/ActiveUserRetriever.js
binh-dam-ibigroup: I would prefer to keep each component's concern separate.

lib/manager/actions/user.js
binh-dam-ibigroup: Yes, thanks, there was one missing check I added in https://github.com/ibi-group/datatools-ui/pull/834/commits/2fe29e2f9f3ecd6229dc3222e7a67ac2cd8fa374. That also fixed flicker on Chrome when logging out.

lib/public/components/UserAccount.js
binh-dam-ibigroup: Sure, updated in https://github.com/ibi-group/datatools-ui/pull/834/commits/a871c71c89865ccd87f4ae423a487d7293684cc1.

lib/common/containers/ActiveUserRetriever.js
```diff
@@ -0,0 +1,41 @@
...
+  // update the user info in the redux state.
+  useEffect(async () => {
+    if (isAuthenticated) {
+      const tokenResponse = await auth0.getAccessTokenSilently({
+        detailedResponse: true
+      })
+      // TRICKY: What we are passing to the backend is the id_token portion of the response.
+      // Unlike the access_token portion, the id_token contains the app_metadata info for the user.
+      receiveTokenAndProfile({ profile, token: tokenResponse.id_token })
+    } else if (!isLoading) {
```
miles-grant-ibigroup, line 34: I think `isLoading` might have some unstable behavior -- I'm seeing the `USER_LOGGED_OUT` redux action firing right after I log in (on Firefox). The result is that the "Learn more about data tools" disappears, but the "Existing users [sign in here](http://localhost:9966/login)." does not disappear!

lib/common/containers/App.js
```diff
@@ -212,12 +123,27 @@ export default class App extends React.Component<AppProps> {
...
+    const routerWithAuth0 = (
+      <Auth0Provider
+        audience=''
+        // Continue to cache tokens in localstorage (speeds up updating login state when refreshing pages).
+        cacheLocation='localstorage'
+        clientId={AUTH0_CLIENT_ID}
+        domain={AUTH0_DOMAIN}
+        onAccessTokenError={this.handleAccessTokenError}
+        onRedirectCallback={this.handleRedirect}
+        // FIXME: This page will briefly flash after login.
```
miles-grant-ibigroup: I haven't been able to log in yet but based on what you said this can probably go!

lib/manager/actions/user.js
```diff
@@ -146,12 +113,36 @@ export function resendEmailConfirmation (user: ManagerUserState): Promise<boolea
...
+        'POST',
+        {
+          client_id: AUTH0_CLIENT_ID,
+          connection: AUTH0_CONNECTION_NAME,
+          email: user.profile.email
+        },
+        true // return the response object itself.
+      )
+    )
+    alert(response.ok
```
miles-grant-ibigroup, line 138: If `ok` is missing from the response will this return false or will it just fail?

miles-grant-ibigroup (changes requested):
Still having problems logging in... I've cleared cache, localStorage, and cookies but still having no luck (Webkit + Firefox)

lib/common/containers/ActiveUserRetriever.js
miles-grant-ibigroup: No errors in the console I should add

lib/manager/actions/user.js
binh-dam-ibigroup: It will treated as if sending the email failed.

lib/common/containers/ActiveUserRetriever.js
binh-dam-ibigroup: The logout action that happens here is equivalent of the one [here](https://github.com/ibi-group/datatools-ui/blob/cc70292c3f9172253e7dbfd477a381d43cad8419/lib/common/user/Auth0Manager.js#L303) and is meant to handle cases when `isAuthenticated` is false (e.g. a login expires) and auth0 is not doing a check in the background for the login status (`isLoading` is false)

miles-grant-ibigroup:
I'm now able to log in, but until I think this should stay blocked until we can confirm that the permissions screen can be removed and that user permissions can be migrated to the new auth0 config.

binh-dam-ibigroup:
> I'm now able to log in, but until I think this should stay blocked until we can confirm that the permissions screen can be removed and that user permissions can be migrated to the new auth0 config.

I did create a new client id for dev purposes, but one thing we can do once this is ready is to update the token algorithms on the current client, so that existing users don't need to be migrated.

miles-grant-ibigroup:
Would this remove the permissions approval screen as well?

miles-grant-ibigroup (approved):

binh-dam-ibigroup:
> Would this remove the permissions approval screen as well?

Per [these docs](https://auth0.com/docs/get-started/applications/confidential-and-public-applications/user-consent-and-third-party-applications), the authorization screen can be skipped, except where `localhost` appears in the auth0 application settings.

lib/common/containers/App.js
```diff
@@ -156,6 +51,19 @@ type AppProps = {
 }
 
 export default class App extends React.Component<AppProps> {
+  handleAccessTokenError = (err: any) => {
+    const newError = new Error('accessToken not received from auth0')
+    console.log(err)
```
daniel-heppner-ibigroup: Should the be a warn or error perhaps?

daniel-heppner-ibigroup (changes requested):
I'm having trouble signing in... Not sure what @miles-grant-ibigroup did to be able to sign in or if I'm doing something wrong. I've updated my configuration. I did notice if I was already signed in on the browser before switching to this branch, then I can't seem to get to the sign in page at all. But in an incognito window I can get to the sign in page, but after signing in, I just get sent back to the main page with the sign in link.

lib/common/components/Login.js
```diff
@@ -1,31 +1,35 @@
...
+          appState: {
+            returnTo: redirectOnSuccess
+          }
+        })
+      } else {
+        // Otherwise just go to the target url.
+        browserHistory.push(redirectOnSuccess)
+      }
+    }
+  }, [isAuthenticated, isLoading, redirectOnSuccess])
```
daniel-heppner-ibigroup, line 34: Is this going to cause the page to redirect any time one of these changes? Could that potentially be a problem?

lib/common/components/Login.js
```diff
@@ -1,31 +1,35 @@
...
+export default function Login ({
+  redirectOnSuccess = window.location.path || '/home'
+}: Props) {
+  const auth0 = useAuth0()
+  const { isAuthenticated, isLoading } = auth0
+  useEffect(() => {
+    if (!isLoading && redirectOnSuccess) {
+      if (!isAuthenticated) {
+        // Show the login page if the user state is not logged in.
+        auth0.loginWithRedirect({
```
daniel-heppner-ibigroup: Extract `loginWithRedirect` above (line 16)

lib/common/components/Login.js
```diff
@@ -1,31 +1,35 @@
...
+          appState: {
+            returnTo: redirectOnSuccess
```
daniel-heppner-ibigroup, line 26: Is this just the regular `auth0` library? I don't think this is the correct argument structure for the redirect url. Perhaps that is why I'm having trouble logging in?
https://auth0.com/docs/libraries/auth0-single-page-app-sdk#login-with-redirect

daniel-heppner-ibigroup: I changed it to redirect_uri as specified in the documentation above, which results in a more helpful auth0 error. 
image
Adding `http://localhost:9966` to the uri causes the error message to change to this. `unauthorized_client: Callback URL mismatch. http://localhost:9966/home/ is not in the list of allowed callback URLs`

lib/common/containers/ActiveUserRetriever.js
```diff
@@ -0,0 +1,41 @@
...
+      logout(auth0)
+    }
+  }, [isAuthenticated, isLoading, profile])
```
daniel-heppner-ibigroup, line 42: This component doesn't seem to be working to grab the user for me. Maybe that's just because I'm not fully getting signed in though. Did it work for you @miles-grant-ibigroup?

daniel-heppner-ibigroup (changes requested):
A few more comments which might be helpful in fixing the bugs....

lib/common/containers/ActiveUserRetriever.js
miles-grant-ibigroup: It took a lot of fiddling with auth0 settings to get right. I think some sort of migration guide/plan is crucial to getting this PR through fully

binh-dam-ibigroup: Actually, what appears to be a migration issue comes from the need to copy datatools permissions for the test client app in Auth0 I created for this PR. When this is merged, the idea is to switch the existing Auth0 app to use the new RS256 algorithm, in which case no user migration will be needed.

lib/common/components/Login.js
binh-dam-ibigroup: The scenario to keep in mind involves a user pasting a URL to a feed version or editor page (e.g. URLs containing `/feed/...`, `/editor/...`). When directly using `redirectUri`, each of the full paths must be whitelisted as callback URLs, which is impractical, whereas with `appState.returnTo`, only the root path needs to be passed to `redirectUri` in the `<Auth0Provider>` and whitelisted in the callback URLs.

lib/common/components/Login.js
binh-dam-ibigroup: Actually the purpose of this component is to redirect ASAP, so nothing happens (and nothing is rendered) until variables change so that the redirect condition is met.

binh-dam-ibigroup: I added the JSDoc about redirecting in https://github.com/ibi-group/datatools-ui/pull/834/commits/0260d14e0fc52e7ebbd8eb1359c13032345feb2b.

lib/common/components/Login.js
binh-dam-ibigroup: Thanks, handled in https://github.com/ibi-group/datatools-ui/pull/834/commits/d1e6d87914721e2a9fef5f5aac11cb0df7e64e01.

lib/common/containers/ActiveUserRetriever.js
binh-dam-ibigroup: auth0-react returns an error message if for some reason it is not able to process the login in the situation that Daniel describes above.  With https://github.com/ibi-group/datatools-ui/pull/834/commits/2e33a20b6f5e07577e55896a86202d7572d682f3, that message will be shown to help with troubleshooting.

lib/common/containers/App.js
binh-dam-ibigroup: Copy pasta... I can't recall why I created that handler in another repo, but it turns out `Auth0Provider` does not have an `onAccessTokenError` callback so I have deleted it in https://github.com/ibi-group/datatools-ui/pull/834/commits/d1e6d87914721e2a9fef5f5aac11cb0df7e64e01...

lib/common/containers/ActiveUserRetriever.js
binh-dam-ibigroup: @daniel-heppner-ibigroup Try again and see if an error message is shown after logging in.

daniel-heppner-ibigroup (approved):
Thank you for the changes and helping me fix the problems I had!

PR #836, opened 2022-08-12
(Flex): Don't send pattern stop defaults to pattern locations

miles-grant-ibigroup:
### Checklist

### Description

Pattern locations CAN have a shape id, pickup type, and drop-off type, however when they have these it can have unintended consequences. Therefore we no longer ship default values for this (and in some cases shipping invalid default values)

As part of this, it is no longer possible to set a pickup or drop-off type for a location. This has to be done via booking rules, which is best practice anyway.

### DIFF
--- a/lib/editor/actions/active.js
+++ b/lib/editor/actions/active.js
@@ -396,8 +396,8 @@ export function saveEntity (
     // Add default vals for component
     const defaults = {}
     if (component === 'route') {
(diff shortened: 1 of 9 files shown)


daniel-heppner-ibigroup (approved):

philip-cline (approved):
This works well for me. I'm still not sure if there may be a better way to define defaults, but for this PR that seems out of scope.

PR #839, opened 2022-08-15
refactor(ExceptionDate): Tweak field width and use ambient font size.

binh-dam-ibigroup:
Enhances layout of the range exception picker from #828!

### DIFF
--- a/lib/style.css
+++ b/lib/style.css
@@ -281,6 +281,11 @@ td.feed-source-info {
   box-shadow: none;
 }
 
+/* Exception date range date picker */
+.entity-details-body .input-group.date .input-group-addon {
+  padding: 6px 4px;
+}
+
 /* Feed version status styling */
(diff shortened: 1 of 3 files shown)


lib/style.css
```diff
@@ -281,6 +281,11 @@ td.feed-source-info {
   box-shadow: none;
 }
 
+/* Exception date range date picker */
+.entity-details-body .input-group.date .input-group-addon {
```
binh-dam-ibigroup: You might need an additional CSS class in `ExceptionDateRange` so that this does not apply to the existing date time picker.

philip-cline: Yep this did need a class name to avoid styling the other date pickers, this should be fixed in https://github.com/ibi-group/datatools-ui/pull/839/commits/1f35f86ca4e64ac41fd21ae8f148a7da54efb222

PR #841, opened 2022-08-17
Fix feed source name editing

philip-cline:
### Checklist

### Description

Currently, when editing the name of a feed source, beginning additions to the name with spaces will cause the spaces to be trimmed as soon as they are entered. This PR resolves that issue.

### DIFF
--- a/lib/manager/components/GeneralSettings.js
+++ b/lib/manager/components/GeneralSettings.js
@@ -40,9 +40,11 @@ type State = {
 export default class GeneralSettings extends Component<Props, State> {
   state = {}
 
-  _onChange = ({target}: SyntheticInputEvent<HTMLInputElement>) => {
+  _onChange = ({target}: SyntheticInputEvent<HTMLInputElement>, key: 'name' | 'url') => {
+    let value
     // Change empty string to null to avoid setting URL to empty string value.
-    const value = target.value === '' ? null : target.value.trim()
+    if (key === 'url') value = target.value === '' ? null : target.value.trim()
+    else value = target.value
     this.setState({[target.name]: value})
   }
 
@@ -130,7 +132,7 @@ export default class GeneralSettings extends Component<Props, State> {
                     value={this._getFormValue('name')}
                     name={'name'}
                     disabled={disabled}
-                    onChange={this._onChange} />
+                    onChange={(e) => this._onChange(e, 'name')} />
                   <InputGroup.Button>
                     <Button
                       // disable if no change or no value (name is required).
@@ -164,7 +166,7 @@ export default class GeneralSettings extends Component<Props, State> {
                   <FormControl
                     disabled={disabled}
                     name={'url'}
-                    onChange={this._onChange}
+                    onChange={(e) => this._onChange(e, 'url')}
                     value={this._getFormValue('url')}
                   />
                   <InputGroup.Button>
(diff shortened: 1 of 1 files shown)


miles-grant-ibigroup (changes requested):
~This fix works but I'm not quite sure why we're making this change? Is there some use case for having a project name start with spaces? It seems like doing so will only cause confusion with no upsides.~

I was editing project titles, not feed source titles. I understand what's happening now. Sorry for the confusion!

lib/manager/components/GeneralSettings.js
```diff
@@ -40,9 +40,11 @@ type State = {
 export default class GeneralSettings extends Component<Props, State> {
   state = {}
 
-  _onChange = ({target}: SyntheticInputEvent<HTMLInputElement>) => {
+  _onChange = ({target}: SyntheticInputEvent<HTMLInputElement>, key: 'name' | 'url') => {
```
miles-grant-ibigroup: You don't need a `key` variable, you should be able to use `target.name` to pull the value from the html tag

lib/manager/components/GeneralSettings.js
```diff
@@ -40,9 +40,11 @@ type State = {
...
+    let value
     // Change empty string to null to avoid setting URL to empty string value.
-    const value = target.value === '' ? null : target.value.trim()
+    if (key === 'url') value = target.value === '' ? null : target.value.trim()
+    else value = target.value
```
miles-grant-ibigroup: Would it not be cleaner to always set `value = target.value` unless it's a url in which case set the trim inside an if block? That way it's one fewer check

lib/manager/components/GeneralSettings.js
```diff
@@ -130,7 +132,7 @@ export default class GeneralSettings extends Component<Props, State> {
                     value={this._getFormValue('name')}
                     name={'name'}
                     disabled={disabled}
-                    onChange={this._onChange} />
+                    onChange={(e) => this._onChange(e, 'name')} />
```
miles-grant-ibigroup: This change can be reverted once we switch to using `target.name`

miles-grant-ibigroup (changes requested):
Fix works great, just a few opportunities for cleanup!

lib/manager/components/GeneralSettings.js
philip-cline: Astounding. Changed in https://github.com/ibi-group/datatools-ui/pull/841/commits/07ed8e52df2b04bd228f7bf2d850f03a41005f9a

lib/manager/components/GeneralSettings.js
philip-cline: Much cleaner. See https://github.com/ibi-group/datatools-ui/pull/841/commits/07ed8e52df2b04bd228f7bf2d850f03a41005f9a

lib/manager/components/GeneralSettings.js
philip-cline: You betcha.

miles-grant-ibigroup (approved):
Wow this is much cleaner! Thanks for this

daniel-heppner-ibigroup (approved):
Seems like before spaces weren't allowed at all. This feels like a `feat: add supports for spaces`.

PR #842, opened 2022-08-18
Editor lock param

binh-dam-ibigroup:
### Checklist

- [na] Any modified or new methods or classes have helpful JSDoc and code is thoroughly commented

### Description

This PR works with and without https://github.com/ibi-group/datatools-server/pull/479, and ~the two can be merged independently.~  EDIT: this PR can be merged independently of the backend one.

This refactor PR simply passes an `item` parameter to the editor lock endpoints. Right now, the `item` parameter is hardcoded to be `gtfs-editor`, although this can/will be changed in the future to accommodate the GTFS+ editor.

### DIFF
--- a/lib/editor/actions/editor.js
+++ b/lib/editor/actions/editor.js
@@ -31,6 +31,8 @@ import type {
 
 import {clearGtfsContent, saveActiveGtfsEntity, setActiveGtfsEntity} from './active'
 
+const LOCK_API_PATH = '/api/editor/secure/lock'
+
 export const updateEntitySort = createAction('UPDATE_ENTITY_SORT')
 
 const createGtfsEntity = createAction(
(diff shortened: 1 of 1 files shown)


miles-grant-ibigroup (approved):
This works great! Thanks for the changes

lib/editor/actions/editor.js
```diff
@@ -31,6 +31,10 @@ import type {
 
 import {clearGtfsContent, saveActiveGtfsEntity, setActiveGtfsEntity} from './active'
 
+const LOCK_API_PATH = '/api/editor/secure/lock'
+/** Identifies what is being locked. */
+const GTFS_EDITOR_LOCK = 'gtfs-editor'
```
daniel-heppner-ibigroup, line 36: The item parameter can basically be whatever, right? As long as the clients always use the same value (like gtfs-editor) then it'll work. So perhaps there should be some enum or const array that limits the possible values for this parameter. (I haven't looked at the GTFS+ locking PR yet, sorry if this is done there)

daniel-heppner-ibigroup (approved):
Looks good to me. Just had some thought about making the possible values for the item, but I suspect it's answered in the next PR.

PR #843, opened 2022-08-22
Don't allow adding new dates until all dates are valid

philip-cline:
### Checklist

### Description

This PR prevents the creation of any new dates until all entered dates are valid (new empty dates are considered invalid).
This PR is currently a WIP and needs to account for multiple dates being edited rather than added.


lib/editor/util/validation.js
```diff
@@ -31,7 +31,7 @@ export function validate (
   entities: ?Array<Entity>,
   entity: ?Entity,
   tableData: EditorTables
-): EditorValidationIssue | false {
+): EditorValidationIssue | false | Array<EditorValidationIssue> {
```
miles-grant-ibigroup, line 34: Is there any way to make this always return an array? I think the single value/array union type will cause more problems than it solves.

lib/editor/components/ScheduleExceptionForm.js
```diff
@@ -122,12 +122,30 @@ export default class ScheduleExceptionForm extends Component<Props> {
     } = activeEntity
 
     // getRangesForDates requires a sorted array. We also sort the zeroes to the end of the array for display purposes.
+    let addDateDisabled = false // Disable adding a new date if we come across any current validation issues.
```
miles-grant-ibigroup: Do we need an explicit variable for this? Since we're using it only in one place can't we just count the number of validation errors inside the `disabled` value?

lib/editor/components/ScheduleExceptionForm.js
```diff
@@ -122,12 +122,30 @@ export default class ScheduleExceptionForm extends Component<Props> {
...
     const {ranges, parsedDates} = getRangesForDates([...dates].sort((a: any, b: any) => {
       if (a === 0) return 1
+      else if (b === 0) return -1
```
miles-grant-ibigroup: `else` is not needed here

lib/editor/components/ScheduleExceptionForm.js
```diff
@@ -122,12 +122,30 @@ export default class ScheduleExceptionForm extends Component<Props> {
...
+      } else {
+        const endIndex = dates.findIndex(d => d === cur.endDate)
+        validationState = this._checkRangeValidation(dateIndex, endIndex)
+        acc.ranges.push({
+          ...cur,
+          validationState
+        })
+      }
+
+      if (validationState === 'error') addDateDisabled = true
```
miles-grant-ibigroup: Writing to global variables from inside a reducer is really bad practice! Good news is that the variable isn't needed so this can be removed

miles-grant-ibigroup: Wow this is a huge expansion of this reducer! I would tear this out and do the validation in a separate loop. Without the diff I don't think I would've understood what this is doing. 

After the reducer (as it was before) you have a ranges array and a parsedDates array. You can then loop over both doing a `.map().filter(not null)` and get 2 clean arrays that you can then use the `length` property of to replace `addDateDisabled`

lib/editor/components/ScheduleExceptionForm.js
```diff
@@ -217,25 +235,22 @@ export default class ScheduleExceptionForm extends Component<Props> {
...
             {ranges && ranges.map((range, index) => {
-              const startDateIndex = dates.findIndex(d => d === range.startDate)
-              const endDateIndex = dates.findIndex(d => d === range.endDate)
               return (
                 <ExceptionDateRange
                   index={index}
                   key={index}
-                  range={range}
-                  validationState={this._checkRangeValidation(startDateIndex, endDateIndex)}
+                  range={{startDate: range.startDate, endDate: range.endDate}}
```
miles-grant-ibigroup: Why was this expansion needed?

miles-grant-ibigroup (changes requested):
A few comments to make this more readable. Haven't tested functionality yet.

lib/editor/util/validation.js
philip-cline: We could, but that would mean modifying the `validate` method to always return an array. Because the method handles a bunch of different validation cases it has a lot of returns and this would be a change in each one of those places. Happy to make that change, just lmk what you think

lib/editor/components/ScheduleExceptionForm.js
philip-cline: We do not, I'll update this

lib/editor/components/ScheduleExceptionForm.js
philip-cline: It wasn't actually, I modified this in https://github.com/ibi-group/datatools-ui/pull/843/commits/bcfe4911d3d622f0ebf397e12c47e6b12df3ea45

lib/editor/components/ScheduleExceptionForm.js
philip-cline: The changes to this file have been reverted and the disabling of the add date btn simplified in https://github.com/ibi-group/datatools-ui/pull/843/commits/bcfe4911d3d622f0ebf397e12c47e6b12df3ea45

philip-cline: Yep, this is removed in https://github.com/ibi-group/datatools-ui/pull/843/commits/bcfe4911d3d622f0ebf397e12c47e6b12df3ea45

lib/editor/components/ScheduleExceptionForm.js
philip-cline: Updated in https://github.com/ibi-group/datatools-ui/pull/843/commits/bcfe4911d3d622f0ebf397e12c47e6b12df3ea45

lib/editor/util/validation.js
miles-grant-ibigroup: Ah makes sense. Let's leave it as is then!

miles-grant-ibigroup (approved):
Much cleaner now and works great! Thanks for fixing this

lib/editor/components/ExceptionDate.js
```diff
@@ -120,7 +123,7 @@ export default class ExceptionDate extends Component<Props, State> {
         <div style={{gridColumn: 'span 5'}}>
           <DateTimeField
             inputProps={{style: inputStyleProps}}
-            key={`date-${index}`}
+            key={uuidv4()}
```
binh-dam-ibigroup: What was the need for using the `key` prop here?

lib/editor/components/ExceptionDateRange.js
```diff
@@ -122,7 +123,7 @@ export default class ExceptionDateRange extends Component<Props, State> {
             className='date-range'
             dateTime={startMoment}
             inputProps={{style: inputStyleProps}}
-            key={`date-${index}-start`}
+            key={uuidv4()}
```
binh-dam-ibigroup: Same remark regarding removing all `key` props.

lib/editor/components/ExceptionDate.js
binh-dam-ibigroup: I would remove all `key` props in this component, they are not part of a loop, and I can't see a need to recreate components in the context of this particular code.

lib/editor/components/ScheduleExceptionForm.js
```diff
@@ -233,7 +234,7 @@ export default class ScheduleExceptionForm extends Component<Props> {
               return (
                 <ExceptionDate
                   date={date}
-                  index={dates.findIndex(d => d === date)}
+                  index={dateIndex}
                   key={index}
```
binh-dam-ibigroup: Maybe it makes sense, in this loop, to use uuids as keys, so that new date pickers are recreated when exceptions are combined (otherwise, if the index is the same but dates at that index change, components are not updated). Same thing when rendering the `ExceptionDateRange` elements above.

lib/editor/components/ExceptionDateRange.js
```diff
@@ -140,7 +141,7 @@ export default class ExceptionDateRange extends Component<Props, State> {
             className='date-range'
             dateTime={range.endDate ? +moment(range.endDate) : +moment(range.startDate).add(1, 'days')}
             inputProps={{style: dateTimeProps}}
-            key={`date-${index}-end`}
+            key={uuidv4()}
             minDate={startMoment && startMoment.clone().startOf('day').add(2, 'days')} // See above, minDate subtracts an extra day.
             onChange={(millis) => this._onRangeUpdate(millis, false)}
             {...dateTimeProps}
```
binh-dam-ibigroup, line 154: One thing we didn't catch in #828: if there is a validation error on a date range, the corresponding message should be shown.

binh-dam-ibigroup (changes requested):
Could you use the `key` prop in the correct places, please?

miles-grant-ibigroup:
E2E tests are passing on this branch!

lib/editor/components/ExceptionDate.js
philip-cline: Sure, removed them in https://github.com/ibi-group/datatools-ui/pull/843/commits/a83521ba6f5a2e66f8642193ca4a3c27efd71908. I believe they were being used to maintain the reference to the DateTimePicker, but moving them up to the loop level appears to work fine.

lib/editor/components/ExceptionDateRange.js
philip-cline: Removed in https://github.com/ibi-group/datatools-ui/pull/843/commits/a83521ba6f5a2e66f8642193ca4a3c27efd71908

lib/editor/components/ScheduleExceptionForm.js
philip-cline: Yes I used the uuids here, this works better than using them at the DateTimePicker level

philip-cline: Updated as of https://github.com/ibi-group/datatools-ui/pull/843/commits/a83521ba6f5a2e66f8642193ca4a3c27efd71908

lib/editor/components/ExceptionDateRange.js
philip-cline: Ranges will now show an error as of https://github.com/ibi-group/datatools-ui/pull/843/commits/a83521ba6f5a2e66f8642193ca4a3c27efd71908

lib/editor/components/ExceptionDateRange.js
```diff
@@ -62,6 +61,9 @@ export default class ExceptionDateRange extends Component<Props, State> {
       // $FlowFixMe: Flow doesn't recognise array filter above.
       dates = updateDates(isStartDate, newRangeBoundary, dates, startMoment, endMoment)
 
+      // We need to sort the dates before update so that indices are maintained for valdiation.
```
binh-dam-ibigroup: Typo "validation"

lib/editor/components/ExceptionDateRange.js
```diff
@@ -62,6 +61,9 @@ export default class ExceptionDateRange extends Component<Props, State> {
...
+      dates.sort((a, b) => moment(a).diff(moment(b)))
```
binh-dam-ibigroup: Lines 65-71 and 90-96 could become a method.

binh-dam-ibigroup (approved):
Changes look good, thanks!

PR #847, opened 2022-08-24
Remove router override for ELB preview

philip-cline:
### Checklist

### Description

Previously, when using a build over wire server the `Preview` button would attach a router name to the URL. However, with an ELB server only the default router is used and thus the router name in the URL causes issues with the preview. This PR removes that router name from the URL in the case that we're using an ELB server


miles-grant-ibigroup (approved):
Have tested this locally and it works great thanks for the changes!

demory (approved):

PR #852, opened 2022-09-09
GTFS+ Editor lock

binh-dam-ibigroup:
### Checklist

### Description

This PR adds lock feature to the GTFS+ editor. The new functionality should work as follows:
- GTFS+ Editor: New behavior: when clicking "Edit GTFS+" from multiple tabs/browsers, the first tab wins, and the "Relock feed" prompt should be shown when other tabs get the focus.
- GTFS Editor: Editor lock behavior remains the same.
- Opening one GTFS Editor and one GTFS+ editor: The lock on one editor should not interfere with the other editor (the behavior remains the same).
- Opening the GTFS editor for the first time after creating a feed: behavior remains the same.

Need feedback on combining the `componentWillReceiveProps` on the `GtfsEditor` and `GtfsPlusEditor`, I have not been able to cleanly separate feed locking from gtfs data fetching (see https://github.com/ibi-group/datatools-ui/compare/editor-lock-param-2...editor-lock-param-2b?expand=1 where it breaks initial load of the GTFS/GTFS+ editor)


lib/gtfsplus/components/GtfsPlusEditor.js
```diff
@@ -135,21 +172,19 @@ export default class GtfsPlusEditor extends Component<Props> {
...
     const activeTable = this.GTFS_PLUS_SPEC.find(t => t.id === activeTableId)
     if (!activeTable) {
       console.warn(`Could not locate GTFS+ table with id=${activeTableId}`)
       return null
     }
 
     return (
-      <ManagerPage ref='page'>
+      <ManagerPage itemToLock={editorActions.GTFS_PLUS_EDITOR_LOCK} ref='page'>
+        <ActiveEditorLockManager
```
miles-grant-ibigroup, line 184: When clicking "re-lock feed" the sidebar disappears! I'm tagging this line since it perhaps is where the issue is coming from, but am not sure what is actually responsible. 

Screen Shot 2022-09-14 at 9 13 44 AM

lib/editor/actions/editor.js
```diff
@@ -31,6 +31,11 @@ import type {
 
 import {clearGtfsContent, saveActiveGtfsEntity, setActiveGtfsEntity} from './active'
 
+const LOCK_API_PATH = '/api/editor/secure/lock'
+// Identifies what is being locked.
+export const GTFS_EDITOR_LOCK = 'gtfs-editor'
+export const GTFS_PLUS_EDITOR_LOCK = 'gtfs-lock-editor'
```
miles-grant-ibigroup: Why is the second string not `gtfs-plus-editor`?

lib/editor/containers/ActiveEditorLockManager.js
```diff
@@ -0,0 +1,118 @@
...
+type Props = {
+  checkLockStatus: typeof editorActions.checkLockStatus,
+  disableLock?: boolean,
+  errorStatus: StatusState,
+  // feedIsLocked: boolean,
+  feedSourceId: string,
+  itemToLock: string,
+  lockEditorFeedSource: typeof editorActions.lockEditorFeedSource,
+  // lockEditorFeedSourceIfNeeded: typeof editorActions.lockEditorFeedSourceIfNeeded,
+  // removeEditorLock: typeof editorActions.removeEditorLock,
```
miles-grant-ibigroup: These can probably go! (there's a few others in this file too)

lib/gtfsplus/components/GtfsPlusField.js
```diff
@@ -140,14 +147,16 @@ export default class GtfsPlusField extends Component<Props> {
           : null
         return (
           <GtfsSearch
+            clearable={false}
+            disabled={disableEditing}
```
miles-grant-ibigroup, line 151: Not sure how I feel about disabling every field individually. Is there no parent component we can replace/hide instead?

miles-grant-ibigroup (changes requested):
I've tested this a bunch and it seems to work really well! A few small clean-ups and one small bug but once those are resolved I think this is definitely good to go!

In terms of combining `componentWillReceiveProps` I don't see much issue with how you've done it here. The duplication present is clearly a result of the way in which the editor lock was built originally. Unless you created some util methods that take a lot of arguments I don't see how this could be much cleaner. It makes sense to me given the context of datatools and works well. 

To me the issue with the locking architecture is less about the locking itself but instead how we manage editor state to begin with

lib/gtfsplus/components/GtfsPlusField.js
binh-dam-ibigroup: I thought we would still let users browse the data even though the editor is locked.

lib/gtfsplus/components/GtfsPlusEditor.js
binh-dam-ibigroup: Did you see any message in the console when that happened? 

This is what I see:
Screen Recording 2022-09-14 at 2 51 32 PM

lib/editor/actions/editor.js
binh-dam-ibigroup: Yikes - fixed in https://github.com/ibi-group/datatools-ui/pull/852/commits/da9d219113c3b89f02471f3f9ab1becc0b1e2fa7.

lib/editor/containers/ActiveEditorLockManager.js
binh-dam-ibigroup: Yes, removed in https://github.com/ibi-group/datatools-ui/pull/852/commits/da9d219113c3b89f02471f3f9ab1becc0b1e2fa7.

lib/gtfsplus/components/GtfsPlusField.js
miles-grant-ibigroup: Yeah that is worth it you're right!

lib/gtfsplus/components/GtfsPlusEditor.js
miles-grant-ibigroup: Apologies, I was on the wrong data tools-server build

miles-grant-ibigroup (approved):
This is working great for me! Any issues I have seem to be longstanding datatools issues rather than issues specifically related to this PR (most notably that you can't refresh the page while in the gtfs editor).

lib/common/components/ManagerPage.js
```diff
@@ -4,18 +4,19 @@ import * as React from 'react'
...
 import { getConfigProperty, isModuleEnabled } from '../util/config'
 
+import ConfirmModal from './ConfirmModal'
+import InfoModal from './InfoModal'
+import SelectFileModal from './SelectFileModal'
+import Title from './Title'
+
 type Props = {
   children?: React.Node,
+  itemToLock?: string,
```
daniel-heppner-ibigroup, line 19: What if we made this string into its own type- an enum that restricts the values to one for the regular editor and one for gtfs+?

lib/editor/actions/editor.js
```diff
@@ -183,15 +188,16 @@ export function stopLockTimer () {
 function maintainEditorLock (
   sessionId: string,
   feedId: string,
+  itemToLock: string,
   timer?: IntervalID
 ) {
   return function (dispatch: dispatchFn, getState: getStateFn) {
-    const url = `/api/editor/secure/lock/${sessionId}?feedId=${feedId}`
+    const url = `${LOCK_API_PATH}/${sessionId}?feedId=${feedId}&item=${itemToLock}`
     // Note: this action is called only when the browser tab/window is visible.
```
daniel-heppner-ibigroup, line 196: I'm having trouble seeing how this ties in to the new component which seems to manage visibility. Is this definitely still true?

daniel-heppner-ibigroup: OK, I'm seeing that the new component disables the timer when the page isn't visible, so I think I get it now!

daniel-heppner-ibigroup (approved):
Just one typing thing that I think maybe could be improved, but I'm not sure if Flow makes enums easy or not.

daniel-heppner-ibigroup (approved):
This is working great for me!

lib/common/components/ManagerPage.js
binh-dam-ibigroup: Right now these are the only two things that can be locked, however if there is impetus to implement collaborative GTFS editing, `itemToLock` could represent any field on any editable entity.

PR #855, opened 2022-09-20
User account types

### LINKED ISSUE #856: Datatools pulls you back to the project dashboard if you view a project before all are loaded.

Datatools navigates back to the home page (dashboard) if a user navigates away from it before its projects finish fetching.

This can create awkward behaviour for the user, who will be forcibly brought back to the project dashboard  if they try to access a project before all feeds are loaded. In the case of a very large project like NYSDOT, this can create a loop of awkward behaviour. 

## Expected behavior

The user should not be brought back to the home page.

1. From the home page, navigate away before the list of projects finishes loading.
2. Observe navigation back to the home page.

NA

Current datatools UI versions.

binh-dam-ibigroup:
### Checklist

- [x] Appropriate branch selected => requires branch `react-17` to take advantage of auth0 features from #834.

### Description

New config settings involved. Fix #856.

This PR adds optional account types to `server.yml` so that for certain account types, a prompt for non-app-admin users to accept license terms is shown until they accept the terms:
- Admins set the user account types under the user admin page when creating or editing users.
   * A default account type can be set to reduce the number of clicks for the admin users, and to limit the fields written in Auth0 for users with such account type.
- The user account type and an external link to the license (if set) are displayed under "My account".
- The license prompt is shown (except for app admins) after logging in (from `/` or after entering a valid datatools URL pointing to a project, feed, editor state, etc.)

Example new `server.yml` settings:
```
modules:
  licensing:
    account_types:
      default:
        name: Standard
        require_terms: false
      trial-period:
        name: Trial Account
        require_terms: true
        # TODO: provide actual terms URL for free accounts
        terms_url: https://mit-license.org/
```

Attention: previously used in this PR, the schema below is no longer valid: 
```
modules:
  licensing:
    enabled: true
    account_types:
    - type: standard
      name: Standard
      default: true
      require_terms: false
    - type: terms-required
      name: Account with terms required
      require_terms: true
      terms_url: https://mit-license.org/
```


miles-grant-ibigroup (commented):
A first overview of the code looks good, but I'm struggling to test this as the dependency chain is getting quite large. I'd suggest blocking this until react-17 is merged and all auth0 changes are merged.

lib/common/containers/ActiveUserRetriever.js
```diff
@@ -41,13 +74,26 @@ const ActiveUserRetriever = ({ logout, receiveTokenAndProfile }) => {
...
 
+const mapStateToProps = (state: AppState) => {
+  return {
+    accountTypes: getAccountTypes(state)
+  }
+}
+
 const mapDispatchToProps = {
   logout: userActions.logout,
   receiveTokenAndProfile: userActions.receiveTokenAndProfile
```
miles-grant-ibigroup, line 96: Not sure if this is a regression or an existing bug, but when changing a user's account type you get logged in as that user!

https://user-images.githubusercontent.com/86619099/192310787-97e35b7f-1305-47d7-b393-ec415225aff3.mov

lib/types/index.js
```diff
@@ -953,6 +959,7 @@ export type UserProfile = {
   name: string,
   nickname: string,
   picture: string,
+  sub: string,
```
miles-grant-ibigroup, line 962: Why was this added?

lib/admin/components/UserSettings.js
```diff
@@ -1,20 +1,34 @@
...
-import type {ManagerUserState} from '../../types/reducers'
+import type { Permission, PermissionType } from '../../common/user/UserPermissions'
+import type { AccountType, Organization, Project } from '../../types'
+import type { ManagerUserState } from '../../types/reducers'
 
 import ProjectAccessSettings from './ProjectAccessSettings'
 
 type Props = {
+  accountType?: ?string,
+  accountTypes: Array<AccountType>,
```
miles-grant-ibigroup: Is there any way we can get this in from redux in this file instead of drilling it all the way in?

miles-grant-ibigroup (changes requested):
Playing around with this functionality seems to work great and I love the interface! Only absolutely blocking issue is a fairly significant security flaw, but it's one that may not be related to this PR.

lib/common/containers/ActiveUserRetriever.js
binh-dam-ibigroup: Yikes, yes, thanks for finding that one. It is introduced by this PR (the behavior does not happen on the deployed dev instance).

binh-dam-ibigroup: Fixed in https://github.com/ibi-group/datatools-ui/pull/855/commits/9e40cc6dd7860fb7d9f47f76de314426e57f1754.

lib/types/index.js
binh-dam-ibigroup: This field contains the user id in profiles returned by `useAuth0`.

lib/admin/components/UserSettings.js
binh-dam-ibigroup: Unfortunately this is not possible without heavier work. `connect`-ing to redux causes existing code to lose access to public methods defined in the class (HOCs do not expose class methods).

miles-grant-ibigroup (approved):
Looks and works great thanks so much for putting this together!

lib/admin/components/UserRow.js
```diff
@@ -204,16 +217,19 @@ export default class UserRow extends Component<Props, State> {
           </Col>
         </Row>
         <ConfirmModal ref='confirm' />
-        { this.state.isEditing
-          ? <UserSettings ref='userSettings'
-            organizations={organizations}
-            projects={projects}
+        {this.state.isEditing && (
+          <UserSettings
+            accountType={userSettings && userSettings.account_type}
```
daniel-heppner-ibigroup, line 218: Would null coalescing not work here?

lib/admin/components/UserSettings.js
daniel-heppner-ibigroup: Maybe using context?

lib/admin/components/UserSettings.js
```diff
@@ -176,8 +209,12 @@ export default class UserSettings extends Component<Props, State> {
     this.setState(stateUpdate)
   }
 
-  _onOrgAdminClicked (evt: SyntheticInputEvent<HTMLInputElement>) {
-    this.setState({orgAdminChecked: evt.target.checked})
+  _handleAccountTypeChanged = (evt: SyntheticInputEvent<HTMLInputElement>) => {
```
daniel-heppner-ibigroup: Name this function following the same style as the others, using "on" instead of "handle"

lib/common/util/user.js
```diff
@@ -2,15 +2,56 @@
...
+export function getAccountTypes (state: AppState): Array<AccountType> {
+  const { appInfo } = state.status
+  const { licensing } = (appInfo && appInfo.config.modules) || {}
+  return (licensing && licensing.enabled && licensing.account_types) || []
+}
+
+/**
+ * Gets the default account type, if any.
+ */
+export function getDefaultAccountType (accountTypes: Array<AccountType>): ?AccountType {
```
daniel-heppner-ibigroup: Could the "default account" be moved up one level in the config so that it's not possible to specify more than one default? Or would it be too difficult to get that config item to this function?

daniel-heppner-ibigroup (approved):
Just a few little nitpicks, but I think it looks good overall! Thank you!

lib/admin/components/UserSettings.js
binh-dam-ibigroup: Updated in https://github.com/ibi-group/datatools-ui/pull/855/commits/877fefb250f08f345ef56b5a35bf733ae16fbd98.

lib/admin/components/UserRow.js
binh-dam-ibigroup: JSX ignores anything that resolves to false too, so the shorter way is to get rid of the `: null` fallback (see https://github.com/ibi-group/datatools-ui/pull/855/commits/877fefb250f08f345ef56b5a35bf733ae16fbd98).

lib/common/util/user.js
binh-dam-ibigroup: Good point, which leads me to question why I picked an array instead of a map to list account types.

lib/admin/components/UserSettings.js
binh-dam-ibigroup: I made the account type selector a connected component in https://github.com/ibi-group/datatools-ui/pull/855/commits/406555e72d66df5c782f929bc11c9e61e50642f4.

lib/common/util/user.js
binh-dam-ibigroup: Thanks for this very good suggestion. Commit https://github.com/ibi-group/datatools-ui/pull/855/commits/406555e72d66df5c782f929bc11c9e61e50642f4 turns `accountTypes` into a map, and the default account type is now denoted with the `default` key, as in below. The `default` key is used in code to seek whether a default account has been set. Let me know what you think @daniel-heppner-ibigroup and @miles-grant-ibigroup:
```
modules:
  licensing:
    account_types:
      default:
        name: Standard
        require_terms: false
      trial-period:
        name: Trial Account
        require_terms: true
        # TODO: provide actual terms URL for free accounts
        terms_url: https://mit-license.org/
```

miles-grant-ibigroup: This is clever! I like it.

daniel-heppner-ibigroup: I like it too! Clever adaptation, I didn't think of this either when I wrote my comment!

PR #858, opened 2022-09-26
Remove maximum time cap on entered timetable string.

philip-cline:
### Checklist

### Description

There is currently a maximum time string of "34:00:00" that can be entered into the timetable editor. Some users, such as very long distance rail carriers, may require times above this. This PR removes any maximum on the times that may be entered into the editor.


miles-grant-ibigroup (changes requested):
Given how chaotic this file is, I won't be requesting too many changes.

One large change that I think might clean things up a lot though is replacing our homegrown parsing with moment? Since we have moment, can't we use use `moment(time, "hh:mm:ss")`?

lib/editor/components/timetable/EditableCell.js
```diff
@@ -208,53 +207,11 @@ export default class EditableCell extends Component<Props, State> {
...
-
-      if (greaterThan24 && momentTime.isValid()) {
-        value += 86400
-      }
-      // Check for valid time and new value
-      const timeIsValid = (time === '' || momentTime.isValid()) && value !== time
-      if (timeIsValid) this._handleSave(value)
+      const duration = moment.duration(data)
+      // $FlowFixMe: flow doesn't know about duration.isValid ensuring valueOf returns a number
+      const value = duration.isValid() ? duration.valueOf() / 1000 : null // ValueOf returns milliseconds
```
miles-grant-ibigroup:
```suggestion
      const value = duration.isValid() && duration.valueOf() / 1000 // ValueOf returns milliseconds
```
This can probably be shortened!

miles-grant-ibigroup (approved):
This looks much much cleaner! I'm so happy. One quick final simplification.

Works great too thanks so much for these changes and wow look at that line diff!

lib/editor/components/timetable/EditableCell.js
miles-grant-ibigroup: You'll probably also have to change the condition to check for `false`

binh-dam-ibigroup (approved):
LGTM

PR #859, opened 2022-09-28
Improve user list

binh-dam-ibigroup:
### Checklist

- [x] Appropriate branch selected => requires merging branch `user-account-types` to `dev` first.

### Description

Blocked by/requires #855. 

This PR improves the list of datatools users in the admin section by converting it to a table and adding info such as account type and last login. App admin badges are now under the account type column. Projects are under the project column The layout of commands while editing a user is changed slightly to accommodate the new user info shown.

Screen Shot 2022-09-28 at 8 38 19 AM

### DIFF
--- a/lib/admin/components/UserRow.js
+++ b/lib/admin/components/UserRow.js
@@ -3,17 +3,18 @@
 import Icon from '@conveyal/woonerf/components/icon'
 import React, { Component } from 'react'
 import { Button, Image, Label as BsLabel } from 'react-bootstrap'
+import { connect } from 'react-redux'
 import uuidv4 from 'uuid/v4'
 
 import * as adminActions from '../actions/admin'
 import ConfirmModal from '../../common/components/ConfirmModal'
 import UserPermissions from '../../common/user/UserPermissions'
 import { getComponentMessages } from '../../common/util/config'
-import { getAccountInfo } from '../../common/util/user'
+import { getAccountTypes, getSettingsFromProfile } from '../../common/util/user'
 import * as feedActions from '../../manager/actions/feeds'
 import * as managerUserActions from '../../manager/actions/user'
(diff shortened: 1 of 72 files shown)


lib/admin/components/UserRow.js
```diff
@@ -154,78 +151,120 @@ export default class UserRow extends Component<Props, State> {
...
+            {!permissions.isApplicationAdmin() && this._getUserPermissionLabel(permissions)}
+            {' '}
+            {this._getOrgLabel(permissions)}
+          </td>
+          {/* When not editing, show the "Last login" cell and
+              show the "actions" column with just the edit button.
+              When editing, hide the "Last login" cell and
+              expand the "actions" column over the space where last login is shown,
+              so that there is enough space to show the delete/save/cancel buttons. */}
+          <td style={isEditing ? { display: 'none' } : null}>
```
miles-grant-ibigroup: Why do we use `display: none` instead of just not rendering the element?

miles-grant-ibigroup (approved):
Looks and works great! Thanks so much. Just one quick question about a small implementation detail, but not blocking.

lib/admin/components/UserRow.js
binh-dam-ibigroup: My intuition was to keep the DOM tree the same instead of adding/removing an element on toggle.

miles-grant-ibigroup: Wouldn't removing the element be the more "react-ey" way to do it? There may be performance implications but I think within datatools these are ignorable.

binh-dam-ibigroup: Updated in https://github.com/ibi-group/datatools-ui/pull/859/commits/60b036c9179c7974ca48bf4c5c7087deb2fe05fa.

philip-cline (approved):

PR #863, opened 2022-09-29
Add Animations to Exception Date List

miles-grant-ibigroup:
### Checklist

### Description

The exception date list could sometimes be a bit jarring when items disappeared. This PR attempts to address this by adding an animation when items are added or removed from the list.

### DIFF
--- a/lib/editor/components/ScheduleExceptionForm.js
+++ b/lib/editor/components/ScheduleExceptionForm.js
@@ -134,6 +134,8 @@ export default class ScheduleExceptionForm extends Component<Props> {
     }, {ranges: [], parsedDates: []})
 
     const hasExemplar = typeof exemplar === 'number'
+
+    const datesAndRangesBothActive = parsedDates && parsedDates.length > 0 && ranges && ranges.length > 0
     return (
       <div>
         <Form>
@@ -224,6 +226,7 @@ export default class ScheduleExceptionForm extends Component<Props> {
               maintainContainerHeight
               typeName={null}
             >
+              {datesAndRangesBothActive && <h6 className='exception-subtitle'>Ranges</h6>}
               {ranges && ranges.map((range, index) => {
                 const startDateIndex = dates.findIndex(d => d === range.startDate)
                 const endDateIndex = dates.findIndex(d => d === range.endDate)
@@ -236,6 +239,7 @@ export default class ScheduleExceptionForm extends Component<Props> {
                     {...this.props}
                   />)
               })}
+              {datesAndRangesBothActive && <h6 className='exception-subtitle'>Dates</h6>}
               {parsedDates && parsedDates.map((date, index) => {
                 const dateIndex = dates.findIndex(d => d === date)
                 return (
--- a/lib/editor/components/ScheduleExceptionForm.js
+++ b/lib/editor/components/ScheduleExceptionForm.js
@@ -135,7 +135,8 @@ export default class ScheduleExceptionForm extends Component<Props> {
 
     const hasExemplar = typeof exemplar === 'number'
 
-    const datesAndRangesBothActive = parsedDates && parsedDates.length > 0 && ranges && ranges.length > 0
+    // If the first date is 0, then we can't be sure it's a date yet!
+    const datesAndRangesBothActive = parsedDates && (parsedDates.length > 0 && parsedDates[0] !== 0) && ranges && ranges.length > 0
     return (
       <div>
         <Form>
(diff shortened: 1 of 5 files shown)


lib/editor/components/ScheduleExceptionForm.js
```diff
@@ -218,29 +220,38 @@ export default class ScheduleExceptionForm extends Component<Props> {
...
-                  date={date}
-                  index={dateIndex}
-                  key={uuidv4()}
-                  validationState={this._checkValidation(`dates-${dateIndex}`)}
-                  {...this.props}
-                />)
-            })}
+            <FlipMove
+              enterAnimation='accordionVertical'
+              leaveAnimation='accordionVertical'
```
philip-cline, line 226: Okay don't hate me but I think I like the elevator animations better?

philip-cline (approved):
Loving this! Slightly prefer the elevator animations, but not a strong preference so will defer to the majority on this one.

binh-dam-ibigroup:
At first, I find all the extra animations very distracting when clicking from one exception to the next... let me look some more.

Screen Recording 2022-10-20 at 12 12 51 PM

lib/editor/components/ScheduleExceptionForm.js
```diff
@@ -218,29 +220,38 @@ export default class ScheduleExceptionForm extends Component<Props> {
...
+              maintainContainerHeight
+              typeName={null}
+            >
+              {datesAndRangesBothActive && <h6 className='exception-subtitle'>Ranges</h6>}
```
binh-dam-ibigroup, line 232: Change `<h6>` to `<p>` (2 instances).

lib/editor/components/ScheduleExceptionForm.js
```diff
@@ -134,6 +134,8 @@ export default class ScheduleExceptionForm extends Component<Props> {
     }, {ranges: [], parsedDates: []})
 
     const hasExemplar = typeof exemplar === 'number'
+
+    const datesAndRangesBothActive = parsedDates && parsedDates.length > 0 && ranges && ranges.length > 0
```
binh-dam-ibigroup: Can adding a new entry (i.e. last parsed date === 0) be excluded from this rule? Adding a new entry in the list of only ranges triggers the rendering of the dates/ranges headings.

lib/editor/components/ScheduleExceptionForm.js
```diff
@@ -218,29 +220,38 @@ export default class ScheduleExceptionForm extends Component<Props> {
...
-                <ExceptionDate
-                  date={date}
-                  index={dateIndex}
-                  key={uuidv4()}
-                  validationState={this._checkValidation(`dates-${dateIndex}`)}
-                  {...this.props}
-                />)
-            })}
+            <FlipMove
+              enterAnimation='accordionVertical'
```
binh-dam-ibigroup, line 225: Insert prop `key={activeEntity.id}` and comment about preventing animations when switching from one item to the next in the list of exceptions.

lib/editor/components/ScheduleExceptionForm.js
binh-dam-ibigroup: I would be in favor of always showing the respective header is there are any ranges/dates under it to help prevent headers from appearing/disappearing too much.

binh-dam-ibigroup (changes requested):
Add the `key` prop as mentioned to address https://github.com/ibi-group/datatools-ui/pull/863#issuecomment-1285826215.

lib/editor/components/ScheduleExceptionForm.js
miles-grant-ibigroup: The reason I was against this is because of the case where it says "on these dates" and then "dates" which I think is really confusing in an interface that is already quite complex

miles-grant-ibigroup: We cannot use `<p>` because we rely on `h6` styles

miles-grant-ibigroup:
Thanks @binh-dam-ibigroup, changes are addressed and things look cleaner.

lib/editor/components/ScheduleExceptionForm.js
binh-dam-ibigroup: Can you move the applicable h6 styles to the `exception-subtitle` CSS?

miles-grant-ibigroup: I don't understand the benefits of duplicating styles like that. Semantically it is correct that these headings are headings, and being level 6 makes it clear that they are not that important, although they are still headers.

binh-dam-ibigroup: These are more similar to control labels than headings.

binh-dam-ibigroup (approved):
I dissent on the `<h6>` thing, but we don't have a policy around that yet.

PR #866, opened 2022-10-14
fix: support deployment summaries, speeding up deployments page

miles-grant-ibigroup:
### Checklist

### Description

Makes use of https://github.com/ibi-group/datatools-server/pull/486. Speeds up the deployment page by not loading in entire deployment object for every entry. Also makes some updates to the mocked server to add a deployment summary endpoint.

Blocked by https://github.com/ibi-group/datatools-server/pull/491


__tests__/test-utils/mock-data/manager.js
```diff
@@ -13,6 +13,23 @@ import type {
 
 let COUNTER = 0
 
+/**
+ * Make a mock deployment given a project and some FeedVersions. This is a
```
philip-cline: I suppose you could say 
```suggestion
 * Make a mock deployment summary given a project and some FeedVersions. This is a
```

lib/manager/components/deployment/DeploymentViewer.js
```diff
@@ -76,6 +76,10 @@ export default class DeploymentViewer extends Component<Props, State> {
   }
 
   componentDidMount () {
+    const { deployment, fetchDeployment } = this.props
+    // If there is no project id, we are dealing with a summary and need to fetch the
```
philip-cline: Should there be a check here to see if there is no project id?

philip-cline (approved):
Nice solution to the project ID issue. Just a few small questions, looks good to me

__tests__/test-utils/mock-data/manager.js
miles-grant-ibigroup: Good catch thank you! Adjusted.

lib/manager/components/deployment/DeploymentViewer.js
miles-grant-ibigroup: Yep thanks for catching this! This comment was outdated.

__tests__/test-utils/mock-data/manager.js
```diff
@@ -13,6 +13,23 @@ import type {
 
 let COUNTER = 0
 
+/**
+ * Make a mock deployment summary given a project and some FeedVersions. This is a
+ * function so that circular references can be defined.
```
binh-dam-ibigroup, line 18: Is there a reason the function args are not used?

lib/manager/reducers/projects.js
```diff
@@ -346,7 +346,10 @@ const projects = (state: ProjectsState = defaultState, action: Action): Projects
       })
     }
     case 'RECEIVE_DEPLOYMENTS': {
-      const {deployments, feedSourceId, projectId} = action.payload
+      const {deployments: deploymentsSummaries, feedSourceId, projectId} = action.payload
+      // Deployment summaries don't include a project id, so we need to add one
+      const deployments = deploymentsSummaries.map(d => ({...d, projectId}))
```
binh-dam-ibigroup: The new `deploymentSummaries` comes with a new field `numberOfFeedVersions` that should now go under the "# of feeds" column in the `DeploymentListItem` component. As written in the PR, the "# of feeds" column in the list of deployments shows N/A until the full deployment object is fetched.

binh-dam-ibigroup (changes requested):
Looks neat, but the "# of feeds" column needs to be populated from the new `DeploymentSummarY` object.

__tests__/test-utils/mock-data/manager.js
miles-grant-ibigroup: Good catch thanks! Parameters have been removed

lib/manager/reducers/projects.js
miles-grant-ibigroup: Thanks for catching this one! Appreciate the heads up about `numberOfFeedVersions`. We now create a fake array.

lib/manager/reducers/projects.js
```diff
@@ -346,7 +346,11 @@ const projects = (state: ProjectsState = defaultState, action: Action): Projects
       })
     }
     case 'RECEIVE_DEPLOYMENTS': {
-      const {deployments, feedSourceId, projectId} = action.payload
+      const {deployments: deploymentsSummaries, feedSourceId, projectId} = action.payload
+      // Deployment summaries don't include a project id, so we need to add one
+      // Deployment summaries don't include feedversions array, so we create a fake array that gets replaced later
+      const deployments = deploymentsSummaries.map(d => ({...d, projectId, feedVersions: Array(d.numberOfFeedVersions)}))
```
binh-dam-ibigroup, line 352: Using an array of `undefined` causes a crash when the user navigates from the Deployments tab back to the Feed Sources tab using the browser's Back button (clicking the Feed Sources tab still works).

The error message is as follows:
```
Uncaught TypeError: Cannot read properties of undefined (reading 'feedSource')
    at version.js:176:28
    at Array.find (<anonymous>)
    at getVersionValidationSummaryByFilterStrategy (version.js:175:67)
    ...
```

binh-dam-ibigroup (changes requested):
Unfortunately the latest change results in a crash when navigating back to the Feed Sources tab, so I'm requesting that to be fixed.

lib/manager/reducers/projects.js
miles-grant-ibigroup: I've fixed this bug, but there is a problem! Since the new "compressed" data doesn't include feed source ids, it's not possible to populate the deployed version column! 

Is there something I'm missing?

binh-dam-ibigroup: Probably an omission on the server side?

miles-grant-ibigroup: Ok so then not something I'm missing then. Do we want to include that though? The list of feeds in each deployment can get quite large especially if there are a lot of deployments with lots of feeds.

I'll schedule a time for us to figure this out offline.

miles-grant-ibigroup:
Have just run the e2e tests locally and they are failing, suggesting we need to fix the feed source columns

miles-grant-ibigroup:
Blocked pending feedSource endpoint updates on the backend.

miles-grant-ibigroup:
Updated to make use of new fields in https://github.com/ibi-group/datatools-server/pull/491. Will wait for e2e tests to pass and then assign to @binh-dam-ibigroup. 

Even though the startDate field is missing, I have defaults set up so it shouldn't impact behavior (aside from the one edge case where the status label will fail to say that a feed which starts in the future starts in the future).

lib/manager/components/version/FeedVersionTabs.js
```diff
@@ -107,8 +106,8 @@ export default class FeedVersionTabs extends Component<Props> {
             <Row>
               {countFields.map(c => {
                 const summary = version.validationSummary
-                const diff = comparedVersion ? (summary[c] - comparedVersion.validationSummary[c]) : 0
-                const value = summary[c]
+                const diff = comparedVersion ? ((summary[c] || 0) - (comparedVersion.validationSummary[c] || 0)) : 0
+                const value = summary[c] || 0
```
binh-dam-ibigroup: Maybe swap these two rows and reuse variable `value`.

binh-dam-ibigroup (approved):
LGTM

lib/manager/components/version/FeedVersionTabs.js
miles-grant-ibigroup: Good catch! Refactored in 15a6f1ad6aa68131ffe4fc4b026bef670d2b1eac

PR #879, opened 2022-11-21
Bugfix: Delete Polygon from Store (Flex)

amy-corson-ibigroup:
### Checklist

### Description

Fixes a bug where if you create a location by drawing a polygon on the map and then delete said polygon, it will not remove the shape from the store (but it will disappear from the map). This will create a "multiple polylines not supported" error if the user attempts to create another polygon. If the user saves the blank map they will see their deleted shape.

Also disables save button when the location has no shapes. 

EDIT: Fixes 2 additional bugs:
- Stops react-draw polyline/polygon control group on map from disappearing even when no shapes exist in location
- If you delete a polygon and re-draw, that polygon will no longer disappear on the map. (bug shown[ here](https://ibigroup-my.sharepoint.com/personal/philip_cline_ibigroup_com/_layouts/15/stream.aspx?id=%2Fpersonal%2Fphilip%5Fcline%5Fibigroup%5Fcom%2FDocuments%2FMicrosoft%20Teams%20Chat%20Files%2FFlex%20new%20polygon%20disappears%2Emov&referrer=Teams%2ETEAMS%2DELECTRON&referrerScenario=p2p%5Fns%2Dbim&ga=1)

### DIFF
--- a/lib/editor/components/map/LocationsLayer.js
+++ b/lib/editor/components/map/LocationsLayer.js
@@ -95,7 +95,7 @@ const LocationsLayer = (props: Props) => {
     const locationShapes = clone(activeEntity.location_shapes)
     const {key} = e
 
-    const newLocationShapes = locationShapes.filter((shapePt) => !(key === shapePt.geometry_id))
+    const newLocationShapes = locationShapes.filter((shapePt) => !(Number(key) === shapePt.geometry_id))
 
     if (!updatingLocationShape) return
     updatingLocationShape({ geometryType, shapes: newLocationShapes })
--- a/lib/editor/components/map/LocationsLayer.js
+++ b/lib/editor/components/map/LocationsLayer.js
@@ -30,7 +30,7 @@ const LocationsLayer = (props: Props) => {
   useEffect(() => {
     // Create our locations shape out of the bulk location shape points
     updateGroupLocationShapePts(groupLocationShapePoints(locationShapes))
-  }, [])
+  }, [locationShapes])
 
   const _onCreated = (e: L.LeafletEvent) => {
     const { layerType: geometryType, layer } = e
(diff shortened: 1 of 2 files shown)


lib/editor/components/map/LocationsLayer.js
```diff
@@ -95,7 +95,7 @@ const LocationsLayer = (props: Props) => {
     const locationShapes = clone(activeEntity.location_shapes)
     const {key} = e
 
-    const newLocationShapes = locationShapes.filter((shapePt) => !(key === shapePt.geometry_id))
+    const newLocationShapes = locationShapes.filter((shapePt) => !(Number(key) === shapePt.geometry_id))
```
miles-grant-ibigroup: Serious nitpick: do you think `parseInt` makes more sense? I didn't even know about `Number` so maybe this is just me not staying up to date with js best practices...

miles-grant-ibigroup (approved):
Can't believe this fix was this easy! Works great in my testing. Thanks for figuring this out!

lib/editor/components/map/LocationsLayer.js
amy-corson-ibigroup: @miles-grant-ibigroup This is a great question, I didn't consider `parseInt`... I _think_ since the goal is type conversion, `Number` might be more appropriate? But in this case they should be functionally the same, so I'm happy to change it! Going off of this: https://stackoverflow.com/questions/4090518/what-is-the-difference-between-parseint-and-number

miles-grant-ibigroup: That was helpful thanks! Let's stick with number then I guess!

amy-corson-ibigroup: Thanks so much!!

philip-cline (approved):
Works stupendously for me, thanks for the fix!

PR #880, opened 2022-11-23
Feat: Zoom to Trip Pattern

amy-corson-ibigroup:
### Checklist

- [x] Appropriate branch selected 

### Description

Adds ability to zoom to a trip pattern on the map. Disables zoom button when trip pattern has empty legs or when there's no trip pattern rendered on the map.

### DIFF
--- a/lib/editor/components/EntityDetailsHeader.js
+++ b/lib/editor/components/EntityDetailsHeader.js
@@ -303,7 +303,6 @@ class EntityDetailsHeader extends Component<Props> {
 const mapStateToProps = (state: AppState, ownProps: Props) => ({
   locations: getLocations(state)
 })
-const mapDispatchToProps = {}
 
-const ActiveEntityDetailsHeader = connect(mapStateToProps, mapDispatchToProps)(EntityDetailsHeader)
+const ActiveEntityDetailsHeader = connect(mapStateToProps)(EntityDetailsHeader)
 export default ActiveEntityDetailsHeader
--- a/lib/editor/components/EntityDetailsHeader.js
+++ b/lib/editor/components/EntityDetailsHeader.js
@@ -2,15 +2,17 @@
 
 import Icon from '@conveyal/woonerf/components/icon'
 import React, {Component} from 'react'
+import {connect} from 'react-redux'
 import {Badge, Button, ButtonGroup, Tooltip, OverlayTrigger, Nav, NavItem} from 'react-bootstrap'
 
 import {getComponentMessages} from '../../common/util/config'
+import {getLocations} from '../selectors'
 import * as activeActions from '../actions/active'
 import * as mapActions from '../actions/map'
 import {getEntityBounds, getEntityName} from '../util/gtfs'
 import {entityIsNew} from '../util/objects'
 import {GTFS_ICONS} from '../util/ui'
-import type {Entity, Feed, GtfsRoute, GtfsStop, Pattern} from '../../types'
+import type {Entity, Feed, GtfsRoute, GtfsLocation, GtfsStop, Pattern} from '../../types'
 import type {MapState} from '../../types/reducers'
 import type {EditorValidationIssue} from '../util/validation'
 
@@ -24,6 +26,7 @@ type Props = {
   editFareRules: boolean,
   entityEdited: boolean,
   feedSource: Feed,
+  locations: Array<GtfsLocation>,
   mapState: MapState,
   resetActiveGtfsEntity: typeof activeActions.resetActiveGtfsEntity,
   saveActiveGtfsEntity: typeof activeActions.saveActiveGtfsEntity,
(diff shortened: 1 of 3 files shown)


lib/editor/components/EntityDetailsHeader.js
```diff
@@ -133,6 +139,11 @@ export default class EntityDetailsHeader extends Component<Props> {
     const zoomDisabled = activeEntity && !subComponent
       ? mapState.target === activeEntity.id
       : mapState.target === subEntity
+    // prevent zooming to empty trip pattern
+    const emptyTripPattern = activeEntity.tripPatterns && activeEntity.tripPatterns.some(pattern => pattern.patternLocations.length === 0 && pattern.patternStops.length === 0)
+    // flex zoom available only when a trip pattern is actually rendered on the map
+    const tripPatternRendered = activeEntity.tripPatterns && !(this.props.params ? this.props.params.subEntityId : false)
```
miles-grant-ibigroup: The extra negation is confusing! I'm not sure if this feels cleaner though... What do you think
```suggestion
    const tripPatternRendered = activeEntity.tripPatterns && (this.props.params ? !this.props.params.subEntityId : true)
```

lib/editor/util/gtfs.js
```diff
@@ -178,7 +178,24 @@ export const generateProps = (component: string, editorState: any): any => {
...
+  function findLocationCoodinates (coordinateArray) {
+    const lats = []
+    const lons = []
+
+    coordinateArray.forEach(point => {
+      // $FlowFixMe flow doesn't understand this but it works
+      lats.push(point.geometry_pt_lat || point.stop_lat)
+      // $FlowFixMe flow doesn't understand this but it works
+      lons.push(point.geometry_pt_lon || point.stop_lon)
+    })
```
miles-grant-ibigroup, line 191: This might be cleaner as a map that maps into two arrays, although I'll leave it up to the second reviewer to decide (especially since this is existing code)

lib/editor/util/gtfs.js
```diff
@@ -220,7 +227,47 @@ export function getEntityBounds (entity: any, stops?: Array<GtfsStop>, offset: n
...
+        }
+
+        // check for flex trip pattern stops, and add to flex coordinates array
+
+        if (stops && pattern.patternStops && pattern.patternStops.length > 0) {
+          flexRouteCoordinates = [
+            ...flexRouteCoordinates,
+            ...stops
+          ]
+        }
```
miles-grant-ibigroup: I think this could be compressed into a single declaration -- both the locations and the stops could be items in a new array and if wrapped in `()` you could add a `|| []` at the end to avoid the null errors.

I realize that's not the best explanation for what I'm trying to describe. We can chat more outside of github

miles-grant-ibigroup (changes requested):
Works really well! Just a little nitpicky cleanup

lib/editor/util/gtfs.js
amy-corson-ibigroup: I was able to make this much shorter, but the `if` condition may be a little long + convoluted so I'm very open to feedback!!

Also may have missed the mark big time, just let me know

lib/editor/util/gtfs.js
```diff
@@ -220,7 +227,40 @@ export function getEntityBounds (entity: any, stops?: Array<GtfsStop>, offset: n
             ...pattern.shape.coordinates.map(c => [c[1], c[0]])
           ]
         }
-      })
+
+        let flexRouteCoordinates = []
```
miles-grant-ibigroup: Does this need to be defined here? Is it used outside the if block?

lib/editor/util/gtfs.js
```diff
@@ -220,7 +227,40 @@ export function getEntityBounds (entity: any, stops?: Array<GtfsStop>, offset: n
...
+          ]
+        }
+
+        // find corners of all location and stop coordinates
+
+        const resolvedFlexCoordinates = flexRouteCoordinates.length > 0 ? findLocationCoodinates(flexRouteCoordinates) : []
+
+        coordinates = [
+          ...coordinates,
+          ...resolvedFlexCoordinates
```
miles-grant-ibigroup: `resolvedFlexCoordinates` can be inlined

miles-grant-ibigroup (approved):
Just 2 more nitpicks and then this is looks good to me! Very clean solution

lib/editor/util/gtfs.js
miles-grant-ibigroup: Looks great!

lib/editor/components/EntityDetailsHeader.js
```diff
@@ -288,3 +299,11 @@ export default class EntityDetailsHeader extends Component<Props> {
     )
   }
 }
+
+const mapStateToProps = (state: AppState, ownProps: Props) => ({
+  locations: getLocations(state)
+})
+const mapDispatchToProps = {}
```
philip-cline: I think you can leave out the mapDispatchToProps if it's not being used. So just 
`connect(mapStateToProps)(EntityDetailsHeader)` below

lib/editor/util/gtfs.js
philip-cline: I kinda wrote the original code here so a little biased but I think this should be fine. 😄

philip-cline (approved):
Looks good, works good. One small nitpick w/ the redux store connection

PR #881, opened 2022-11-29
fix(AddPatternStopDropdown): Allow duplicate locations in patterns

philip-cline:
### Checklist

### Description

The current flex specification has intra-zonal trips defined as two stop times with the same location. Currently we were not allowing duplicate locations in a pattern.


miles-grant-ibigroup (approved):
Just tested wonderful fix thanks! My only nitpick would be double checking if the `stop_id` check is needed in the method but better safety than sorry especially in datatools

amy-corson-ibigroup (approved):

PR #886, opened 2022-12-09
Stop refresh on Trip ID Error

amy-corson-ibigroup:
### Checklist

- [x] Appropriate branch selected

### Description

This PR seeks to address an issue where users who were saving a trip id previously held by a different trip were getting an error that asked them to reload the page and lose their changes.
-  Returns an error message that can be closed instead of prompting refresh
-  Stop state refresh on error
-  Fixes a small error where field could not be reset to its original value after error

### DIFF
--- a/lib/common/actions/index.js
+++ b/lib/common/actions/index.js
@@ -131,6 +131,9 @@ function getErrorMessageFromJson (
     }
     // re-assign error message after it gets used in the detail.
     errorMessage = json.message || JSON.stringify(json)
+    if (json.detail.includes('Key field must be unique')) {
+      action = 'DEFAULT'
+    }
   }
   console.error(errorMessage)
   return { action, detail, message: errorMessage }
--- a/lib/common/actions/index.js
+++ b/lib/common/actions/index.js
@@ -131,7 +131,7 @@ function getErrorMessageFromJson (
     }
     // re-assign error message after it gets used in the detail.
     errorMessage = json.message || JSON.stringify(json)
-    if (json.detail.includes('Key field must be unique')) {
+    if (json.detail.includes('conflicts with an existing trip id')) {
       action = 'DEFAULT'
     }
   }
--- a/lib/editor/actions/trip.js
+++ b/lib/editor/actions/trip.js
@@ -180,7 +180,10 @@ export function saveTripsForCalendar (
         })
     }))
       .then(trips => {
-        dispatch(fetchTripsForCalendar(feedId, pattern, calendarId, true))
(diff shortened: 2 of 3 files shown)


lib/common/actions/index.js
```diff
@@ -131,6 +131,9 @@ function getErrorMessageFromJson (
     }
     // re-assign error message after it gets used in the detail.
     errorMessage = json.message || JSON.stringify(json)
+    if (json.detail.includes('Key field must be unique')) {
```
philip-cline: I believe the detail of the JSON response has changed with the new backend PR (Server PR #[489](https://github.com/ibi-group/datatools-server/pull/489))to be in the format of `"Trip id 2 conflicts with an existing trip id."`.

philip-cline: This check here will need to be updated to avoid the reload button appearing. String matching also feels a little brittle but should be fine since this is a temp fix

philip-cline (commented):
I can test again fully once the condition for the new backend PR is in there

philip-cline (approved):
Sorry for the delay in review, but this looks good to me! Thanks for checking against the new error.

lib/common/actions/index.js
```diff
@@ -131,6 +131,9 @@ function getErrorMessageFromJson (
     }
     // re-assign error message after it gets used in the detail.
     errorMessage = json.message || JSON.stringify(json)
+    if (json.detail.includes('conflicts with an existing trip id')) {
+      action = 'DEFAULT'
```
miles-grant-ibigroup, line 135: Can we add a `console.warn` or similar letting us know that we did a slightly strange thing?

lib/editor/actions/trip.js
```diff
@@ -180,7 +180,10 @@ export function saveTripsForCalendar (
         })
     }))
       .then(trips => {
-        dispatch(fetchTripsForCalendar(feedId, pattern, calendarId, true))
+        // if error in saving, state will not refresh so user does not lose work
+        if (!trips.includes(undefined)) {
+          dispatch(fetchTripsForCalendar(feedId, pattern, calendarId, true))
+        }
```
miles-grant-ibigroup, line 186: Same here -- could we possibly add an `else` and a console log saying that we haven't updated because we have an invalid trip!

miles-grant-ibigroup (approved):
Both such a simple and complex fix at the same time. Thanks for figuring this out! Approved pending the `console.warn`s

amy-corson-ibigroup:
Thanks everyone!

PR #889, opened 2022-12-16
Default flex pickup dropoff types

binh-dam-ibigroup:
### Checklist

- [x] Appropriate branch selected => this PR is for flex features and is thus against `dev-flex`.

### Description

This PR reinstates the pickup/dropoff type selector for flex stops in the route editor, and adjusts the defaults to 2 (must phone agency) for new flex stops and stops where that info is not provided.

The behavior when loading a feed or editor state where flex stops have invalid pickup/dropoff types is not defined until the values are changed. Let me know if that needs to be addressed with this PR.


lib/editor/components/pattern/PatternStopCard.js
```diff
@@ -522,7 +551,7 @@ class PatternStopContents extends Component<Props, State> {
             </FormGroup>
           </Col>
         </Row>
-        {feedSource.flexUIFeaturesEnabled && pickupDropOffRow}
+        {feedSource.flexUIFeaturesEnabled && this._renderPickupDropOffTypes(false)}
```
philip-cline: This condition should be inverted or placed in a ternary with the condition below right? Right now this displays nothing if `flexUIFeaturesEnabled` is false: 
image

lib/editor/components/pattern/PatternStopCard.js
```diff
@@ -522,7 +551,7 @@ class PatternStopContents extends Component<Props, State> {
...
         {feedSource.flexUIFeaturesEnabled && (
```
philip-cline: I think based on this we would also never show continuous pickup and dropoff fields for a normal stop as well right? This seems like a legacy issue

lib/editor/components/pattern/PatternStopCard.js
```diff
@@ -544,7 +573,7 @@ class PatternStopContents extends Component<Props, State> {
                 selectType='continuousDropOff'
                 shouldHaveDisabledOption
                 title='Indicates whether a rider can alight from the transit vehicle at any point along the vehicle\u2019s travel path.'
-                value={(patternStop.continuousDropOff || patternStop.continuousDropOff === 0) ? patternStop.continuousDropOff : 1}
+                value={numberOrDefault(patternStop.continuousDropOff, 1)}
```
philip-cline: I'm also never seeing these fields rendered for flex or non-flex feeds

philip-cline (changes requested):
Couple things that don't seem to be working

lib/editor/components/pattern/PatternStopCard.js
binh-dam-ibigroup: Is `feedSource.flexUIFeaturesEnabled` true in your case?

binh-dam-ibigroup: I can change it if that's needed for non-flex feeds.

lib/editor/components/pattern/PatternStopCard.js
binh-dam-ibigroup: No, that function does not control rendering of continous pickup/dropoff fields.

lib/editor/components/pattern/PatternStopCard.js
binh-dam-ibigroup: You will have to show me the config you use. I see this for regular stops in a flex-enabled feed:

image

lib/editor/components/pattern/PatternStopCard.js
philip-cline: If we test a non-flex feed on the flex branch, it will be false and we'll only see the screenshot above. We'll need it to still look the same as in current PROD: 
image

lib/editor/components/pattern/PatternStopCard.js
philip-cline: Right, but it is avoiding rendering the pickup and dropoff for a non-flex feed in the flex branch right?

lib/editor/components/pattern/PatternStopCard.js
philip-cline: This is fine, now sure what was going on with my config here.

lib/editor/components/pattern/PatternStopCard.js
binh-dam-ibigroup: Okay, I think I addressed this in https://github.com/ibi-group/datatools-ui/pull/889/commits/88d32d5ae870cccaea5a15e71b1b258463b026b8. The only difference is that, for a non-flex route in a flex-enabled feed, the booking rules fields will still appear (and default with blank).

lib/editor/components/pattern/PatternStopCard.js
binh-dam-ibigroup: See https://github.com/ibi-group/datatools-ui/pull/889#discussion_r1053364100.

philip-cline (approved):
Looks good now, thanks for updating!

lib/editor/components/pattern/PatternStopCard.js
```diff
@@ -280,6 +285,20 @@ class PatternStopCard extends Component<Props> {
   }
 }
 
+/**
+ * Whether a pattern halt is a flex location/location group.
+ */
+function isFlexLocationOrLocationGroup (patternStop: PatternHalt) {
+  return !!patternHaltIsLocation(patternStop) || !!patternHaltIsLocationGroup(patternStop)
+}
```
miles-grant-ibigroup: Could this entire method not be replaced with a negative check for patternStop? I feel this is cleaner than adding an entirely new method

miles-grant-ibigroup (approved):
Can't comment as much as Phil on functionality but seems to work well for me!

Just one small nitpick, but looks great otherwise

lib/editor/components/pattern/PatternStopCard.js
binh-dam-ibigroup: Good find, this cleans up more code!

PR #894, opened 2023-01-06
disable normalize stops when locations are present

amy-corson-ibigroup:
### Checklist

### Description

Removes the Normalize Stop Times button on flex routes. Also removes the associated tips that reference a button that does not exist.

### DIFF
--- a/lib/editor/components/pattern/CalculateDefaultTimesForm.js
+++ b/lib/editor/components/pattern/CalculateDefaultTimesForm.js
@@ -17,6 +17,7 @@ import * as tripPatternActions from '../../actions/tripPattern'
 import MinuteSecondInput from '../MinuteSecondInput'
 import {getDistanceScaleFactor, straightLineDistancesBetweenStopAnchors} from '../../util/map'
 import type {ControlPoint, Pattern} from '../../../types'
+import { activePatternHasLocations } from '../../util/location'
 
 import NormalizeStopTimesTip from './NormalizeStopTimesTip'
 
@@ -96,7 +97,6 @@ export default class CalculateDefaultTimesForm extends Component<Props, State> {
 
   render () {
     const { activePattern } = this.props
-    const activePatternHasLocations = activePattern.patternLocations.length > 0 || activePattern.patternLocationGroups.length > 0
 
     return (
       <Form>
@@ -129,7 +129,7 @@ export default class CalculateDefaultTimesForm extends Component<Props, State> {
           </InputGroup.Button>
         </InputGroup>
         <br />
-        {!activePatternHasLocations &&
+        {!activePatternHasLocations(activePattern) &&
           <Well style={{padding: '8px'}}>
             <NormalizeStopTimesTip />
           </Well>
--- a/lib/editor/components/pattern/CalculateDefaultTimesForm.js
+++ b/lib/editor/components/pattern/CalculateDefaultTimesForm.js
@@ -95,6 +95,9 @@ export default class CalculateDefaultTimesForm extends Component<Props, State> {
     this.setState({speed: +evt.target.value})
 
   render () {
+    const { activePattern } = this.props
+    const activePatternHasLocations = activePattern.patternLocations.length > 0
+
     return (
       <Form>
         <FormGroup className={`col-xs-4`} bsSize='small'>
@@ -126,9 +129,11 @@ export default class CalculateDefaultTimesForm extends Component<Props, State> {
           </InputGroup.Button>
         </InputGroup>
         <br />
-        <Well style={{padding: '8px'}}>
-          <NormalizeStopTimesTip />
-        </Well>
+        {!activePatternHasLocations &&
+          <Well style={{padding: '8px'}}>
+            <NormalizeStopTimesTip />
+          </Well>
+        }
       </Form>
     )
   }
--- a/lib/editor/components/pattern/CalculateDefaultTimesForm.js
+++ b/lib/editor/components/pattern/CalculateDefaultTimesForm.js
@@ -96,7 +96,7 @@ export default class CalculateDefaultTimesForm extends Component<Props, State> {
 
   render () {
     const { activePattern } = this.props
-    const activePatternHasLocations = activePattern.patternLocations.length > 0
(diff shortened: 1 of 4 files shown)


lib/editor/components/pattern/CalculateDefaultTimesForm.js
```diff
@@ -95,6 +95,9 @@ export default class CalculateDefaultTimesForm extends Component<Props, State> {
     this.setState({speed: +evt.target.value})
 
   render () {
+    const { activePattern } = this.props
+    const activePatternHasLocations = activePattern.patternLocations.length > 0
```
philip-cline: We should check for location groups too here, because those would require the same new logic to actually implement `Normalize Stop Times`

lib/editor/components/pattern/PatternStopCard.js
```diff
@@ -472,6 +473,7 @@ class PatternStopContents extends Component<Props, State> {
     const pickupBookingRuleId = getGtfsSpecField('stop_time', 'pickup_booking_rule_id')
     const dropOffBookingRuleId = getGtfsSpecField('stop_time', 'drop_off_booking_rule_id')
     const isStop = patternHaltIsStop(patternStop)
+    const activePatternHasLocations = activePattern.patternLocations.length > 0
```
philip-cline: Same thing here

lib/editor/components/pattern/PatternStopsPanel.js
```diff
@@ -114,6 +114,7 @@ export default class PatternStopsPanel extends Component<Props, State> {
           activePattern.patternLocationGroups.length > 0)
 
     const patternHaltCount = (activePattern.patternStops.length || 0) + (activePattern.patternLocations.length || 0) + (activePattern.patternLocationGroups.length || 0)
+    const patternHasLocations = activePattern.patternLocations.length > 0
```
philip-cline: Same thing here

lib/editor/components/pattern/PatternStopsPanel.js
```diff
@@ -128,15 +129,17 @@ export default class PatternStopsPanel extends Component<Props, State> {
...
-          <Button
-            onClick={this._clickNormalizeStopTimes}
-            style={{marginBottom: '5px'}}
-            className='pull-right'
-            block
-            bsSize='small'>
-            <Icon type='clock-o' />{' '}
-            Normalize stop times
-          </Button>
+          {!patternHasLocations &&
```
philip-cline: I'm not sure if I prefer disabling to fully removing the button in this case, I'll defer to the second reviewer.

philip-cline (changes requested):
Couple changes to account for LocationGroups

lib/editor/components/pattern/CalculateDefaultTimesForm.js
amy-corson-ibigroup: Very good point thank you!!!

lib/editor/components/pattern/PatternStopsPanel.js
amy-corson-ibigroup: I think I lean removing because perhaps disabling creates the impression there's an error on the page the user needs to fix before they're allowed to use the button? But I'm also happy to defer!!

philip-cline (approved):
Nice, thanks for the fix! That's a good point, I think I agree on removing rather than disabling now.

lib/editor/components/pattern/CalculateDefaultTimesForm.js
```diff
@@ -95,6 +95,9 @@ export default class CalculateDefaultTimesForm extends Component<Props, State> {
     this.setState({speed: +evt.target.value})
 
   render () {
+    const { activePattern } = this.props
+    const activePatternHasLocations = activePattern.patternLocations.length > 0 || activePattern.patternLocationGroups.length > 0
```
miles-grant-ibigroup: PatternStopsPanel.js already has a method `patternHasStops`. I would extract this into a utils file and then use that (negated) here!

lib/editor/components/pattern/PatternStopCard.js
```diff
@@ -472,6 +473,7 @@ class PatternStopContents extends Component<Props, State> {
     const pickupBookingRuleId = getGtfsSpecField('stop_time', 'pickup_booking_rule_id')
     const dropOffBookingRuleId = getGtfsSpecField('stop_time', 'drop_off_booking_rule_id')
     const isStop = patternHaltIsStop(patternStop)
+    const activePatternHasLocations = activePattern.patternLocations.length > 0 || activePattern.patternLocationGroups.length > 0
```
miles-grant-ibigroup: You can re-use that extracted method here too!

lib/editor/components/pattern/PatternStopsPanel.js
```diff
@@ -114,6 +114,7 @@ export default class PatternStopsPanel extends Component<Props, State> {
           activePattern.patternLocationGroups.length > 0)
 
     const patternHaltCount = (activePattern.patternStops.length || 0) + (activePattern.patternLocations.length || 0) + (activePattern.patternLocationGroups.length || 0)
+    const activePatternHasLocations = activePattern.patternLocations.length > 0 || activePattern.patternLocationGroups.length > 0
```
miles-grant-ibigroup: And here too!

miles-grant-ibigroup (approved):
Looks good! Just one small nitpick about re-using `patternHasStops`

lib/editor/components/pattern/PatternStopsPanel.js
miles-grant-ibigroup: Let's get rid of the button!

lib/editor/components/pattern/CalculateDefaultTimesForm.js
amy-corson-ibigroup: We discussed this partly off of github, but `patternHasStops` isn't quite right for this purpose, and `patternHaltIsStop` can check individual stops/locations/location groups, but not full patterns. I went ahead and extracted `activePatternHasLocations` into a util!

PR #895, opened 2023-01-11
Don't Refetch feed source list when opening project

miles-grant-ibigroup:
### Checklist

- [x] e2e tests are all passing _(remove this if not merging to master)_

### Description

Currently we re-fetch the entire feed source list when opening a project, even though we already have all that information downloaded. Really the true fix here is to look at our data insertion pipeline. For now, this PR adds a check that prevents the re-fetch from happening if we already have the needed data. 

I would be open to a more holistic solution if anyone has one in mind!

### DIFF
--- a/lib/manager/actions/feeds.js
+++ b/lib/manager/actions/feeds.js
@@ -50,12 +50,6 @@ const FS_URL = '/api/manager/secure/feedsource'
  */
 export function fetchProjectFeeds (projectId: string) {
   return function (dispatch: dispatchFn, getState: getStateFn) {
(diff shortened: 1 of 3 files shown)


philip-cline (approved):
I can't think of a better fix personally, and this seems clean enough especially since it will make a big difference in load times in the short term. 

You're on a roll with DT PRs rn (rocket ship emoji)

daniel-heppner-ibigroup (approved):
I'm too out of touch with DT right now to consider better approaches, but this works. Thanks!

PR #901, opened 2023-01-27
Fetch project if saving a new feed

philip-cline:
### Checklist

### Description

Previous changes introduced a bug where if creating and saving a new feed source in a project, the project would not be fetched again, caused a loading error. This PR resolves that, which also resolves the current e2e issues.

### DIFF
--- a/lib/manager/actions/projects.js
+++ b/lib/manager/actions/projects.js
@@ -109,7 +109,7 @@ export function fetchProject (projectId: string, unsecure: ?boolean) {
 
 export function fetchProjectWithFeeds (projectId: string, unsecure: ?boolean) {
   return function (dispatch: dispatchFn, getState: getStateFn) {
-    // Don't fetch project or feeds if project and feeds are already present
+    // Don't fetch project or feeds if project and feeds are already present and the feed sources and projects aren't being actively saved
     const state = getState()
     const project = state.projects.all.find(p => p.id === projectId)
     const isSaving = state.status.message && state.status.saving
--- a/lib/manager/actions/projects.js
+++ b/lib/manager/actions/projects.js
@@ -112,7 +112,7 @@ export function fetchProjectWithFeeds (projectId: string, unsecure: ?boolean) {
     // Don't fetch project or feeds if project and feeds are already present
     const state = getState()
     const project = state.projects.all.find(p => p.id === projectId)
-    const isSaving = state.status.message && state.status.message.includes('Saving')
+    const isSaving = state.status.message && state.status.saving
     if (!isSaving && project && project.feedSources && project.feedSourceCount && project.feedSources.length <= project.feedSourceCount) {
       return
     }
--- a/lib/manager/actions/projects.js
+++ b/lib/manager/actions/projects.js
@@ -6,16 +6,11 @@ import {browserHistory} from 'react-router'
 import {createVoidPayloadAction, secureFetch} from '../../common/actions'
 import {getConfigProperty, isModuleEnabled} from '../../common/util/config'
 import {deploymentsEnabledAndAccessAllowedForProject} from '../../common/util/permissions'
-import {fetchProjectDeployments} from './deployments'
-import {fetchProjectFeeds} from './feeds'
-import {handleJobResponse} from './status'
 import {getActiveProject} from '../selectors'
 import {
   projectHasAtLeastOneDeployment,
   projectHasAtLeastOneFeedWithAPublishedVersion
 } from '../util'
-import {setVisibilitySearchText} from './visibilityFilter'
-
 import type {Project} from '../../types'
 import type {
   dispatchFn,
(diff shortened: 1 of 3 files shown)


miles-grant-ibigroup (approved):
Thanks for fixing this!

lib/manager/actions/projects.js
```diff
@@ -112,7 +112,8 @@ export function fetchProjectWithFeeds (projectId: string, unsecure: ?boolean) {
     // Don't fetch project or feeds if project and feeds are already present
```
binh-dam-ibigroup: Can you update this comment?

lib/manager/actions/projects.js
```diff
@@ -112,7 +112,8 @@ export function fetchProjectWithFeeds (projectId: string, unsecure: ?boolean) {
...
     const state = getState()
     const project = state.projects.all.find(p => p.id === projectId)
-    if (project && project.feedSources && project.feedSourceCount && project.feedSources.length === project.feedSourceCount) {
+    const isSaving = state.status.message && state.status.message.includes('Saving')
```
miles-grant-ibigroup: I just realized this doesn't support i18n, so this fix won't work in Polish or German. Maybe we should set another redux variable when saving?

binh-dam-ibigroup: Indeed. That would also clarify which action this behavior should apply.

binh-dam-ibigroup (changes requested):
What Miles said.

lib/manager/reducers/status.js
```diff
@@ -71,7 +72,7 @@ const config = (state: StatusState = defaultState, action: Action): StatusState
     case 'UPDATING_SERVER':
     case 'FETCHING_ALL_JOBS':
     case 'REQUEST_RTD_ALERTS':
-      return {...state, message: messages(action.type)}
+      return {...state, message: messages(action.type), isSaving: action.type === 'SAVING_FEEDSOURCE'}
```
binh-dam-ibigroup: Is that `isSaving` status getting unset after the the feed source is saved, or through another action above (and which one(s))?

philip-cline: The parameter will get set to false after the feed source is saved by any subsequent action in the case list. 
image

lib/manager/actions/projects.js
philip-cline: This is updated in [c8f54e8](https://github.com/ibi-group/datatools-ui/pull/901/commits/c8f54e8d224542f56e53b71eb69dbeda954af84a)

lib/manager/reducers/status.js
miles-grant-ibigroup: What's setting it back to false? Do you think this could be the source of the ci failures

binh-dam-ibigroup: From looking at the code, add to the comment on line 75 that `REQUESTING_FEED_SOURCE` and `REQUESTING_PROJECT` set the `isSaving` state back to false, which is what we want.

binh-dam-ibigroup (approved):
Add a comment regarding resetting the `isSaving` state per https://github.com/ibi-group/datatools-ui/pull/901#discussion_r1093461538.

lib/manager/reducers/status.js
philip-cline: Updated in [ba34342](https://github.com/ibi-group/datatools-ui/pull/901/commits/ba343422d596b78b3aa21de98f7fc734ca288ff4)

miles-grant-ibigroup (changes requested):
Looks like the tests are failing randomly -- I'd remove the extra `wait`s and instead try to fix the problem those waits are solving.

miles-grant-ibigroup (approved):
Looks like things are working now!

philip-cline:
This has gone back and forth a bit since your last review @binh-dam-ibigroup, so just requesting another review

lib/manager/actions/projects.js
binh-dam-ibigroup: Is the update to the comment you made gone?

lib/manager/reducers/status.js
```diff
@@ -71,7 +73,7 @@ const config = (state: StatusState = defaultState, action: Action): StatusState
     case 'UPDATING_SERVER':
     case 'FETCHING_ALL_JOBS':
     case 'REQUEST_RTD_ALERTS':
-      return {...state, message: messages(action.type)}
+      return {...state, saving: true, message: messages(action.type)}
```
binh-dam-ibigroup, line 76: Add comment regarding what sets `saving` back to `false`. Also, a previous commit was setting `saving` back to `false` but I'm not seeing that, meaning the logic is incomplete.

binh-dam-ibigroup (changes requested):
Looks strange. Something happened and some previous valid changes were lost.

miles-grant-ibigroup:
@binh-dam-ibigroup all changes should be back now!

binh-dam-ibigroup:
> @binh-dam-ibigroup all changes should be back now!

Great, thanks!

binh-dam-ibigroup (approved):

PR #902, opened 2023-01-31
Allow use of route override for continuous pickup and dropoff

philip-cline:
### Checklist

### Description

Currently, default (unselected) `continuous_pickup` and `dropoff` values are stored as "1 - Not Available" in the stop_times table. However, the `routes.continuous_pickup` and `dropoff` values will not be able to override these values if "1" is specified. Rather, the column should be blank if unselected as outlined in the specification so that the route settings may override this. 

This PR adds a new option to the `PickupDropoffSelect` component which I think might be controversial: 
image

However, this option makes it clear that before selecting a value, the behaviour remains the default. I'm open to reworking the text here if we think it would improve the UI.


lib/editor/components/pattern/PatternStopCard.js
```diff
@@ -68,6 +68,10 @@ type State = {
 }
 
 const pickupDropoffOptions = [
+  {
+    value: '',
+    text: '(default)'
```
binh-dam-ibigroup: I like the idea. Maybe say "(Inherit from route)" or something (and add i18n), and perhaps show in the `text` field what the route default is.

binh-dam-ibigroup: Sort attributes.

philip-cline: I was avoiding adding "Inherit from route" or etc. because this component is shared with `Pickup_type` and `dropoff_type`, and those values do not have route level defaults. However, they do have default values if nothing is selected, that's why I chose the more generic language.

miles-grant-ibigroup: Could we maybe add a prop that changes the string? I think `(default)` is very confusing

philip-cline: Sure thing 😁

philip-cline: Thoughts on the following? @binh-dam-ibigroup @miles-grant-ibigroup 
for pickup/dropoff: 
image

for continuous: 
image

philip-cline: To me we could maybe remove the unwieldy `Inherit from route` portion by providing more help text where the user selects the route data itself

philip-cline: Ok thoughts on the following? @binh-dam-ibigroup @miles-grant-ibigroup 
image

image

image

philip-cline: To me we could remove the unwieldy `Inherit from route` portion by adding more help text where the user enters that data in the route menu

philip-cline: Okay updated per our call - this should be ready for review @binh-dam-ibigroup

philip-cline: Updated in [8f9dc88](https://github.com/ibi-group/datatools-ui/pull/902/commits/8f9dc88408c8b43db8ef249b85dde1ddfe753cf6)

lib/editor/components/pattern/PatternStopCard.js
```diff
@@ -95,6 +102,7 @@ const PickupDropoffSelect = (props: PickupDropoffSelectProps) => {
     selectType,
     shouldHaveDisabledOption,
     title,
+    type,
```
binh-dam-ibigroup: Can `selectType` be reused instead?

binh-dam-ibigroup (approved):
Nice work! (See one refactoring suggestion.)

lib/editor/components/pattern/PatternStopCard.js
philip-cline: Great point, I reused that prop in https://github.com/ibi-group/datatools-ui/pull/902/commits/d82eb3bc186433bacdc85571d8fa62f5636b3bdf

lib/editor/components/pattern/PatternStopCard.js
```diff
@@ -117,7 +123,11 @@ const PickupDropoffSelect = (props: PickupDropoffSelectProps) => {
         value={value}
       >
         {pickupDropoffOptions.map(o => (
-          <option key={o.value} value={o.value}>{o.text}</option>
+          <option key={o.value} value={o.value}>
+            {o.value !== '' ? o.text : selectType.includes('continuous')
```
miles-grant-ibigroup, line 127: Is this reading from the gtfs value? Otherwise this will only work in English won't it

miles-grant-ibigroup (changes requested):
Much cleaner now thank you! Just one quick question about getting the labeling right but other than that looks great thanks for the changes!

lib/editor/components/pattern/PatternStopCard.js
philip-cline: `pickupDropoffOptions` is a constant defined in this file (... should it really be, probably not) and I internationalized it in an earlier commit, so those `o.text` values are translated. 
image

philip-cline: This also just reminded me that there's probably a reckoning somewhere down the road with the `translations.txt` file in the spec ....

miles-grant-ibigroup: I meant if `selectType` was guaranteed to contain the word `continuous`, but I think that's a gtfs value right? The other fix... might be out of scope here

philip-cline: Oh yeah for the `selectType.contains()` portion, that is an internal prop so should be ok to use english right?

philip-cline: This also made me think this could be a bit brittle, and typing this prop a bit more could help. (https://github.com/ibi-group/datatools-ui/pull/902/commits/17236207fa0366765c3dfd8bc9083f660f50ed8d)

miles-grant-ibigroup (approved):
Even cleaner! Thanks very much. Looks like the e2e tests are failing but once those are resolved I think this should be good to go!

PR #905, opened 2023-02-07
Populate pattern pickupType/dropOffType into new stop time

binh-dam-ibigroup:
### Checklist

- [na] Any modified or new methods or classes have helpful JSDoc and code is thoroughly commented

### Description

In the timetable editor, when adding a new stop time to a pattern (*not* duplicating an existing one),
the value in fields `pickupType` and `dropOffType` are incorrectly left as null.
This PR populates these fields from the pattern stop as desired.


br648 (approved):
All good. Can confirm that the stop_times table is being populated with pickup and drop off types when stop times are first created. These values are also present in downloaded snapshots.

philip-cline (approved):
wow good catch.

PR #890, opened 2022-12-20
Add stop headsigns

philip-cline:
### Checklist

- [x] Appropriate branch selected _(all PRs must first be merged to `dev` before they can be merged to `master`)_
- [x] Any modified or new methods or classes have helpful JSDoc and code is thoroughly commented
- [x] The description lists all applicable issues this PR seeks to resolve
- [x] The description lists any configuration setting(s) that differ from the default settings
- [x] All tests and CI builds passing

### Description

This PR adds support for stop_headsigns, which can be created at the PatternStop level. This PR requires backend PR #:  https://github.com/conveyal/gtfs-lib/pull/369

PR #921, opened 2023-03-03
Write stop headsigns for new trips

philip-cline:
### Checklist

### Description

Forgot to write stop_headsigns to new trips with recent additions.


br648 (approved):
All good.

miles-grant-ibigroup (approved):
Good catch! Isn't there a third file though that gets updated when we do something like this?

philip-cline:
In the original `Stop_headsign` [PR](https://github.com/ibi-group/datatools-ui/pull/890/files) we added the field to the graphQL `patternStopFields` and then also to the method that creates a patternStop, I believe those should be the other places where the field is needed.

PR #926, opened 2023-03-09
Map

changtung:
I don't know if there are any tests to pass for this app, so if tests are no passing, please let me know about it and ill solve something. 

this branch is functionality that You choose between OSM map and MAPBOX map in env.yml. 
it was done on some older version however i see git shows this can be automatically merged. 
Looking forward for an answer from You:)


miles-grant-ibigroup:
Thanks for your contribution! The lint step is currently failing but you should be able to fix this by running `yarn lint --fix`. We will take a look at if this implementation is our preferred way of adding this feature. But we agree it's important and thank you again for your fix!

changtung:
fixed with lint. now success.
image

changtung:
IT this E2E test working with auth0 client id and password ? I am asking becuse we do not want to submit those details to github repository. How to ommit this if yes @miles-grant-ibigroup ?

miles-grant-ibigroup:
We use GitHub secrets for these. They only get inserted on PRs we open. We might open a new PR for your branch so that they run. We'll keep you posted! Thanks

miles-grant-ibigroup:
Closed in favor of #931. @changtung please let me know if the new PR works for you!

PR #930, opened 2023-03-14
Don't increment cumulativeTravelTime for an intra zonal segment

philip-cline:
### Checklist

### Description

Intrazonal stop times segments should now be effectively duplicates. We avoid incrementing the cumulative travel time in this PR to make this the case. ✨

### DIFF
--- a/lib/editor/components/timetable/TimetableEditor.js
+++ b/lib/editor/components/timetable/TimetableEditor.js
@@ -219,7 +219,7 @@ export default class TimetableEditor extends Component<Props, State> {
           // $FlowFixMe Flow doesn't recognize our check for stopId property?
           cumulativeTravelTime += +halt.flexDefaultTravelTime
 
-          const haltId = getPatternHaltId(halt)
+          const haltId = getPatternHaltId(halt) // Handles check for Location vs. LocationGroup
           const nextHalt = i + 1 !== patternHalts.length ? patternHalts[i + 1] : null
 
           objectPath.set(newRow, `stopTimes.${i}.stopId`, haltId)
--- a/lib/editor/components/timetable/TimetableEditor.js
+++ b/lib/editor/components/timetable/TimetableEditor.js
@@ -224,14 +224,17 @@ export default class TimetableEditor extends Component<Props, State> {
           const nextHalt = i + 1 !== patternHalts.length ? patternHalts[i + 1] : null
           let nextHaltId
           if (nextHalt) {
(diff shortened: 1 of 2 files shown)


lib/editor/components/timetable/TimetableEditor.js
```diff
@@ -220,11 +220,22 @@ export default class TimetableEditor extends Component<Props, State> {
           cumulativeTravelTime += +halt.flexDefaultTravelTime
 
           // $FlowFixMe Flow doesn't recognize our check for stopId property?
-          objectPath.set(newRow, `stopTimes.${i}.stopId`, patternHaltIsLocation(halt) ? halt.locationId : halt.locationGroupId)
+          const haltId = patternHaltIsLocation(halt) ? halt.locationId : halt.locationGroupId
+          const nextHalt = i + 1 !== patternHalts.length ? patternHalts[i + 1] : null
```
miles-grant-ibigroup, line 223: Shouldn't the check be i + 2? Otherwise won't the last halt not be included

lib/editor/components/timetable/TimetableEditor.js
```diff
@@ -220,11 +220,22 @@ export default class TimetableEditor extends Component<Props, State> {
...
+          let nextHaltId
```
miles-grant-ibigroup: Instead of storing the id separately, can't we just create/use a method that returns the id of a halt no matter the halt type?

miles-grant-ibigroup (changes requested):
Thanks for fixing this issue that's been on my todo list for a long time! Just a few cleanup opportunities

lib/editor/components/timetable/TimetableEditor.js
philip-cline: we're checking the nextHalt, so we want to stop when we hit the current halt at the end of the array, so i = length - 1, because that's where there's no "nextHalt"

lib/editor/components/timetable/TimetableEditor.js
philip-cline: I think I agree, https://github.com/ibi-group/datatools-ui/pull/930/commits/e75153f1e1e68b66f7467976b34869a461ec3462 looks a lot cleaner

miles-grant-ibigroup (approved):
Working great, thanks for the changes!

amy-corson-ibigroup (approved):
Looks good to me!!

PR #936, opened 2023-04-03
Add location_id to front end location shapes

philip-cline:
### Checklist

### Description

Regression where we wouldn't include `location_id` when editing and saving a polygon.

### DIFF
--- a/lib/editor/components/map/LocationsLayer.js
+++ b/lib/editor/components/map/LocationsLayer.js
@@ -84,7 +84,7 @@ const LocationsLayer = (props: Props) => {
         geometry_pt_lat: coord.lat,
         geometry_pt_lon: coord.lng,
         id: index,
-        location_id: activeEntity.location_id
+        location_id: activeEntity.location_id // location_id is required within each location_shape point on the backend.
       }
     ))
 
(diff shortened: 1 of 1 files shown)


amy-corson-ibigroup (approved):
Gorgeous!

lib/editor/components/map/LocationsLayer.js
```diff
@@ -83,7 +83,8 @@ const LocationsLayer = (props: Props) => {
         geometry_id: key,
         geometry_pt_lat: coord.lat,
         geometry_pt_lon: coord.lng,
-        id: index
+        id: index,
+        location_id: activeEntity.location_id
```
miles-grant-ibigroup: Wait could you possibly add a comment/explain here why this is needed? Why are we doing this

miles-grant-ibigroup (approved):
All good pending explanatory comment. I have a feeling it might be a multi-line...

lib/editor/components/map/LocationsLayer.js
philip-cline: The backend requires the location_id in each location_shape. We should really be just doing that work on the backend I suppose

philip-cline: See https://github.com/ibi-group/datatools-ui/pull/936/commits/aaf328f3c513c59337fdaad41937e336d4843faa and let me know if that clears it up for you :)

miles-grant-ibigroup:
Looks great thanks! I think keeping it in the front end is fine -- better to pass it than to have to fish it out, no?

philip-cline:
I agree. Easier to finagle on FE.

PR #940, opened 2023-04-12
Don't Create Location Shape Points on Map Drag

miles-grant-ibigroup:
### Checklist

### Description

Leaflet draw has a bug where when dragging the map, it receives both a touch and a click event, causing the map to both pan while a new point is added to a location shape. This behavior is confusing for users. The solution is to disable the leaflet touch event handler when a touch device is not present.

### DIFF
--- a/lib/editor/components/map/EditorMap.js
+++ b/lib/editor/components/map/EditorMap.js
@@ -125,12 +125,12 @@ export default class EditorMap extends Component<Props, State> {
 
     // From https://github.com/Leaflet/Leaflet.draw/issues/695#issuecomment-281663243
     // Prevents creating a new leaflet draw point on drag
-    var originalOnTouch = L.Draw.Polyline.prototype._onTouch;
-  	L.Draw.Polyline.prototype._onTouch = function (e) {
-      if (e.originalEvent.pointerType != "mouse") {
-        return originalOnTouch.call(this, e);
+    var originalOnTouch = L.Draw.Polyline.prototype._onTouch
+    L.Draw.Polyline.prototype._onTouch = function (e) {
+      if (e.originalEvent.pointerType !== 'mouse') {
+        return originalOnTouch.call(this, e)
       }
-   };
+    }
   }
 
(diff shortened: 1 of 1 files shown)


philip-cline (approved):
Wow amazing fix!!!! i can't believe it turned out to be this easy.

lib/editor/components/map/EditorMap.js
```diff
@@ -121,6 +122,15 @@ export default class EditorMap extends Component<Props, State> {
   componentDidMount () {
     window.addEventListener('resize', this._onResize)
     this.setState({willMount: false})
+
+    // From https://github.com/Leaflet/Leaflet.draw/issues/695#issuecomment-281663243
+    // Prevents creating a new leaflet draw point on drag
+    var originalOnTouch = L.Draw.Polyline.prototype._onTouch
```
binh-dam-ibigroup:
```suggestion
    const originalOnTouch = L.Draw.Polyline.prototype._onTouch
```

binh-dam-ibigroup: This is touching the prototype of a global object, therefore, I think this should be moved outside of the component code and run once for all. (There could be an undesired side effect the second time a map mounts.)

binh-dam-ibigroup (commented):
What do you think about moving the prototype change outside of the component?

binh-dam-ibigroup (changes requested):
I'm going to recommend moving the code that changes `L.Draw.Polyline.prototype._onTouch` outside of the component so it runs only once per browser load.

lib/editor/components/map/EditorMap.js
philip-cline: Stuck it at the top of the file in https://github.com/ibi-group/datatools-ui/pull/940/commits/97525ab78f826293bc4ce997de0c8d5e4ba175f6. Let me know what you think

binh-dam-ibigroup (approved):
Latest change looks good, thanks.

PR #943, opened 2023-04-27
Add links to patterns for stop deletion

philip-cline:
### Checklist

### Description

Link a user to the pattern that uses a specific stop when trying to delete it for faster (and much more usable) deletion work flow

### DIFF
--- a/lib/common/actions/index.js
+++ b/lib/common/actions/index.js
@@ -145,7 +145,6 @@ function getErrorMessageFromJson (
           const [patternId, routeId] = pattern.split('-')
           return {patternId, routeId}
         })
-        // $FlowFixMe
         detail = <PatternLinkErrorMessage detail={detail} patterns={patterns} />
       }
     }
--- a/lib/common/actions/index.js
+++ b/lib/common/actions/index.js
@@ -1,6 +1,7 @@
 // @flow
 
 import fetch from 'isomorphic-fetch'
+import React from 'react'
 
 import {GTFS_GRAPHQL_PREFIX} from '../../common/constants'
 import {getComponentMessages} from '../../common/util/config'
@@ -131,12 +132,54 @@ function getErrorMessageFromJson (
     }
     // re-assign error message after it gets used in the detail.
     errorMessage = json.message || JSON.stringify(json)
+    if (json.detail && json.detail.includes('Referenced pattern IDs:')) {
+      // Custom formatting to link to pattern stops to remove if so desired.
+      // Pull the string of patternIds
+      const patternSplit = json.detail.split(/\[(.*?)\]/) // Match text within square brackets (our comma separated list of patternIds)
+      detail = patternSplit[0]
+      const patternsMatch = patternSplit[1]
+      if (patternsMatch) {
+        const patterns = patternsMatch.split(',')
+        // $FlowFixMe
+        detail = <PatternLinkErrorMessage detail={detail} patterns={patterns} />
+      }
+    }
     if (json.detail && json.detail.includes('conflicts with an existing trip id')) {
       action = 'DEFAULT'
     }
   }
-  console.error(errorMessage)
   return { action, detail, message: errorMessage }
+}
+
+const PatternLinkErrorMessage = (props) => {
+  const url = window.location.href
+  const basePath = url.split('edit/')[0] // Remove the current component from the URL
+  const {detail, patterns} = props
+
+  // Limit the list of patterns to 5 to not overwhelm users
+  const unDisplayedPatterns = patterns.length - 5
+  const limitedPatterns = patterns.slice(0, 5)
+
+  // $FlowFixMe
+  return <div>
+    {detail}
+    {limitedPatterns.map((pattern, index) => {
+      if (index > 5) return
+      // Pattern contains route and pattern data in format {pattern-route}
+      pattern = pattern.slice(1, -1) // Remove curly braces
+      const [patternId, routeId] = pattern.split('-')
+
+      const patternPath = `edit/route/${routeId}/trippattern/${patternId}`
+      return (
+        <li style={{listStyle: 'none'}}>
+          <a href={basePath + patternPath}>
+            {`Pattern ${patternId}`}
+          </a>
+        </li>
+      )
+    })}
+    {unDisplayedPatterns > 0 && `... and ${unDisplayedPatterns} other patterns`}
+  </div>
 }
 
 function graphQLErrorsToString (errors: Array<{locations: any, message: string}>): Array<string> {
--- a/lib/common/actions/index.js
+++ b/lib/common/actions/index.js
@@ -132,9 +132,9 @@ function getErrorMessageFromJson (
     }
     // re-assign error message after it gets used in the detail.
     errorMessage = json.message || JSON.stringify(json)
+    // This section (L135 - L146) is a hack to display links to patterns referenced by a stop to delete.
+    // This should be replaced with a better solution such as a new endpoint for this information.
     if (json.detail && json.detail.includes('Referenced pattern IDs:')) {
-      // Custom formatting to link to pattern stops to remove if so desired.
-      // Pull the string of patternIds
       const patternSplit = json.detail.split(/\[(.*?)\]/) // Match text within square brackets (our comma separated list of patternIds)
       detail = patternSplit[0]
       const patternsMatch = patternSplit[1]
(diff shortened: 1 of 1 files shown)


lib/common/actions/index.js
```diff
@@ -131,14 +132,56 @@ function getErrorMessageFromJson (
     }
     // re-assign error message after it gets used in the detail.
     errorMessage = json.message || JSON.stringify(json)
+    if (json.detail && json.detail.includes('Referenced pattern IDs:')) {
```
miles-grant-ibigroup: You're essentially creating your own text-based protocol, which is a massive hack. This is fine for now, but we definitely need to add a comment here (including line numbers) that this needs to go and be replaced with a better solution, namely a new endpoint that returns exactly this information in JSON format

lib/common/actions/index.js
```diff
@@ -131,14 +132,56 @@ function getErrorMessageFromJson (
...
 }
 
+const PatternLinkErrorMessage = (props) => {
+  const url = window.location.href
+  const basePath = url.split('edit/')[0] // Remove the current component from the URL
+  const {detail, patterns} = props
+
+  // Limit the list of patterns to 5 to not overwhelm users
+  const unDisplayedPatterns = patterns.length - 5
+  const limitedPatterns = patterns.slice(0, 5)
```
miles-grant-ibigroup: I would not do this

lib/common/actions/index.js
```diff
@@ -131,14 +132,56 @@ function getErrorMessageFromJson (
...
+
+  // $FlowFixMe
+  return <div>
+    {detail}
+    {limitedPatterns.map((pattern, index) => {
+      if (index > 5) return
```
miles-grant-ibigroup: We don't need this check twice

lib/common/actions/index.js
```diff
@@ -131,14 +132,56 @@ function getErrorMessageFromJson (
...
+    }
     if (json.detail && json.detail.includes('conflicts with an existing trip id')) {
       action = 'DEFAULT'
     }
   }
-  console.error(errorMessage)
   return { action, detail, message: errorMessage }
 }
 
+const PatternLinkErrorMessage = (props) => {
```
miles-grant-ibigroup, line 165: Again, I would add a comment here saying that this entire block needs to be removed when the better solution is put in place. I would reference this PR as well so we remember what we were thinking

lib/common/actions/index.js
```diff
@@ -131,14 +132,56 @@ function getErrorMessageFromJson (
...
+        const patterns = patternsMatch.split(',')
+        // $FlowFixMe
+        detail = <PatternLinkErrorMessage detail={detail} patterns={patterns} />
+      }
+    }
     if (json.detail && json.detail.includes('conflicts with an existing trip id')) {
       action = 'DEFAULT'
     }
   }
-  console.error(errorMessage)
```
miles-grant-ibigroup: Why did you remove this? I found it helpful to have the errors in the console

miles-grant-ibigroup (changes requested):
This is a fine solution on a time crunch, but we need something cleaner in place for the long run

lib/common/actions/index.js
philip-cline: Agreed.

lib/common/actions/index.js
philip-cline: You wouldn't limit the number of patterns shown ? Or just wouldn't slice?

miles-grant-ibigroup: For the proper implementation we could maybe do a scrollable list? For now since this is a duct tape solution I would say show them all

philip-cline: We could probably do a scrollable list within this PR

philip-cline: Ok I added some scrolling, maybe doesn't look the best but we can expand that too in the future.

philip-cline: Indeed. Removed in https://github.com/ibi-group/datatools-ui/pull/943/commits/5a027d784a5ab84f546bc645f448d4771f77e9bc

lib/common/actions/index.js
philip-cline: Comment added. Having a component in here itself kind of seems like a sign of the end times.

lib/common/actions/index.js
philip-cline: Sorry, accident. Restored in https://github.com/ibi-group/datatools-ui/pull/943/commits/5a027d784a5ab84f546bc645f448d4771f77e9bc

lib/common/actions/index.js
```diff
@@ -148,37 +148,48 @@ function getErrorMessageFromJson (
...
+      borderWidth: '2px',
+      maxHeight: '150px',
+      margin: '10px 0px',
+      overflow: 'auto',
+      padding: '5px 10px'
+    }}>
+      {patterns.map((pattern, index) => {
+        // Pattern contains route and pattern data in format {pattern-route}
+        pattern = pattern.slice(1, -1) // Remove curly braces
+        const [patternId, routeId] = pattern.split('-')
```
miles-grant-ibigroup: Just to future-proof this, would it be possible to move these lines out of this component and pass it in as props instead?

lib/common/actions/index.js
```diff
@@ -132,9 +132,9 @@ function getErrorMessageFromJson (
     }
     // re-assign error message after it gets used in the detail.
     errorMessage = json.message || JSON.stringify(json)
+    // This section (L135 - L146) is a hack to display links to patterns referenced by a stop to delete.
```
miles-grant-ibigroup, line 136: Can we add a TODO: REMOVE HACK above this line? Just to make it clear this has to go

miles-grant-ibigroup (changes requested):
Almost there! Just a few nitpicks

lib/common/actions/index.js
philip-cline: ur wish is my command https://github.com/ibi-group/datatools-ui/pull/943/commits/1203494986f3560476ac9251715238cb5c4a9fc1

lib/common/actions/index.js
philip-cline: Nothing I would love more: https://github.com/ibi-group/datatools-ui/pull/943/commits/1203494986f3560476ac9251715238cb5c4a9fc1

lib/common/actions/index.js
```diff
@@ -131,6 +132,23 @@ function getErrorMessageFromJson (
...
+      const patternSplit = json.detail.split(/\[(.*?)\]/) // Match text within square brackets (our comma separated list of patternIds)
+      detail = patternSplit[0]
+      const patternsMatch = patternSplit[1]
+      if (patternsMatch) {
+        const patterns = patternsMatch.split(',').map(pattern => {
+          pattern = pattern.slice(1, -1) // Remove curly braces
+          const [patternId, routeId] = pattern.split('-')
+          return {patternId, routeId}
+        })
+        // $FlowFixMe
```
miles-grant-ibigroup: It feels terrifying to use a fake patterns object like this -- maybe we should add a comment? Or fix the flow types

miles-grant-ibigroup (approved):
Works great! Thanks for all the changes

lib/common/actions/index.js
philip-cline: umm https://github.com/ibi-group/datatools-ui/pull/943/commits/f9f2ba5e447b30a1dfcdd59e7eee31bdeaa62964

lib/common/actions/index.js
```diff
@@ -131,6 +132,22 @@ function getErrorMessageFromJson (
     }
     // re-assign error message after it gets used in the detail.
     errorMessage = json.message || JSON.stringify(json)
+    // TODO: REMOVE HACK
+    // This section (L135 - L146) is a hack to display links to patterns referenced by a stop to delete.
+    // This should be replaced with a better solution such as a new endpoint for this information.
+    if (json.detail && json.detail.includes('Referenced pattern IDs:')) {
```
binh-dam-ibigroup: The searched string is inconsistent with the one defined at https://github.com/conveyal/gtfs-lib/blob/30754ba741ee414fc1778837160d91e12694eddf/src/main/java/com/conveyal/gtfs/loader/JdbcTableWriter.java#L1616, so this will never show the links.
```suggestion
    if (json.detail && json.detail.includes('Referenced patterns:')) {
```

lib/common/actions/index.js
```diff
@@ -139,6 +156,44 @@ function getErrorMessageFromJson (
...
+      maxHeight: '150px',
+      margin: '10px 0px',
+      overflow: 'auto',
+      padding: '5px 10px'
+    }}>
+      {patterns.map((pattern, index) => {
+        const patternPath = `edit/route/${pattern.routeId}/trippattern/${pattern.patternId}`
+        return (
+          <li style={{listStyle: 'none', margin: '3px 0px'}}>
+            <a href={basePath + patternPath}>
```
binh-dam-ibigroup: I'd be curious if the link href could be a relative URL.

binh-dam-ibigroup (approved):
Approving subject to making the referenced patterns string consistent with what is sent from gtfs-lib.

lib/common/actions/index.js
philip-cline: Good catch the two PRs were out of sync. Updated in https://github.com/ibi-group/datatools-ui/pull/943/commits/78fc07f63988a31c5f525297b54e2db1f33d2916

lib/common/actions/index.js
philip-cline: Great idea! I updated it to be a relative URL in https://github.com/ibi-group/datatools-ui/pull/943/commits/78fc07f63988a31c5f525297b54e2db1f33d2916

PR #946, opened 2023-05-02
update router config gui to support OTP2 router-config.json

miles-grant-ibigroup:
### Checklist

### Description

Datatools has a cool router-config gui. Unfortunately, it was unusable because it didn't support any OTP2 parameters. Given that it's entirely out of use at the moment, this PR permanently shifts it to use OTP2 parameters instead. The exposed parameters do not cover every case, but they do cover most of them.

Needs https://github.com/ibi-group/datatools-server/pull/533

### DIFF
--- a/i18n/english.yml
+++ b/i18n/english.yml
@@ -150,7 +150,7 @@ components:
       title: Router Config
       updaters:
         $index:
-          defaultAgencyId: Default agency ID
+          feedId: Feed ID
           frequencySec: Frequency (in seconds)
           sourceType: Source type
           type: Type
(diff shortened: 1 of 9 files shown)


i18n/english.yml
```diff
@@ -539,7 +539,7 @@ components:
       title: Router Config
       updaters:
         $index:
-          defaultAgencyId: Default agency ID
+          feedId: Feed ID
```
philip-cline, line 548: Kinda makes you wonder why we have so many duplicated strings in these files.

i18n/english.yml
```diff
@@ -150,7 +150,7 @@ components:
...
```
philip-cline, line 156: Need to add the other strings for new fields: 
image

philip-cline (approved):
Works well for me, just some missing translations.

i18n/english.yml
miles-grant-ibigroup: Yes it's a bit scary isn't it

amy-corson-ibigroup (approved):

PR #949, opened 2023-05-05
Update shape_dist_traveled after stop removal

### LINKED ISSUE #941: Shape_dist_traveled is not adjusted when removing stop from pattern
## Observed behavior

If you remove a stop from the beginning of the trip pattern using the UI, the `shape_dist_traveled` values will not be adjusted in `stop_times.txt`, even though the shape_dist_traveled values in `shapes.txt` do get updated. This results in a mismatch between the two values.

## Expected behavior

Values in `stop_times.txt::shape_dist_traveled` should get adjusted. The `shape_dist_traveled` of the new first stop should be 0, instead of its distance before the truncation.

1. Create a trip pattern with several stops.
2. Create a trip using the trip pattern.
3. Remove stops from the beginning of the trip pattern. (E.g. remove the first two stops.)
4. Create a new snapshot and download the GTFS.

In the resulting GTFS:
 - `shapes.txt` will be filled out correctly, the first shape point will have 0 as `shape_dist_traveled`.
 - Howerver in `stop_times.txt` stops will have an incorrect `shape_dist_traveled value`. 

None.

## Version of datatools-server and datatools-ui if applicable (exact commit hash or branch name)

datatools-ui version: `1be32bec`

philip-cline:
### Checklist

### Description

Fix #941 Previously, when a stop was removed from a pattern, the relevant shape_dist_traveled values were not updated. This PR adjusts the removal process to update all shape_dist_traveled values and save the activeGTFSEntity after update, so that change are reflected in an immediate export.

### DIFF
--- a/lib/editor/actions/map/stopStrategies.js
+++ b/lib/editor/actions/map/stopStrategies.js
@@ -595,7 +595,7 @@ const generateSegmentAndInsert = async (segments: Array<Coordinates>, fromPoint:
  * These values must also be updated in the patternStops which will be used to create the shapeDistTraveled
  * values in stop_times.txt
  */
-const updateShapeDistTraveled = (controlPoints: Array<ControlPoint | null>, segments: Array<Coordinate>, patternStops: Array<PatternStop>): void => {
+export const updateShapeDistTraveled = (controlPoints: Array<ControlPoint | null>, segments: Array<Coordinate>, patternStops: Array<PatternStop>): void => {
   // $FlowFixMe
   const stopControlPoints = getControlPointsForStops(controlPoints)
 
--- a/lib/editor/actions/map/stopStrategies.js
+++ b/lib/editor/actions/map/stopStrategies.js
@@ -532,6 +532,7 @@ export function removeStopFromPattern (pattern: Pattern, stop: GtfsStop, index: 
       const {avoidMotorways, followStreets} = getState().editor.editSettings.present
       let result
       try {
+        // $FlowFixMe: Flow does not recognize controlpoints within returned Promise type
         result = await recalculateShape({
           avoidMotorways,
           controlPoints: clonedControlPoints,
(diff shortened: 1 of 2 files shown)


binh-dam-ibigroup (commented):
Not seeing the changes to `shape_dist_traveled` after deleting the first two stops on a pattern and taking/downloading a snapshot. Did I miss anything @philip-cline?

philip-cline:
@binh-dam-ibigroup nope, that's a good catch I was only handling cases where control points are left but I updated to include first stop deletion

lib/editor/actions/map/stopStrategies.js
```diff
@@ -527,10 +527,12 @@ export function removeStopFromPattern (pattern: Pattern, stop: GtfsStop, index:
     if (!shapePoints || shapePoints.length === 0) {
       // If pattern has no shape points, don't attempt to refactor pattern shape
       console.log('pattern coordinates do not exist')
```
binh-dam-ibigroup, line 529: Are these comment and console statements still relevant?

binh-dam-ibigroup (approved):
Seems to work now, thanks for the changes.

lib/editor/actions/map/stopStrategies.js
philip-cline: It's still relevant bc we're not calling `updatePatternGeometry`, only splicing & then saving the patternStops

lib/editor/actions/map/stopStrategies.js
```diff
@@ -550,14 +552,16 @@ export function removeStopFromPattern (pattern: Pattern, stop: GtfsStop, index:
         dispatch(setErrorMessage({message: `Could not remove stop from pattern:`}))
         return
       }
+      patternStops.splice(index, 1)
```
miles-grant-ibigroup, line 555: Why is this extracted from being run after the `if/else` block to running both inside the if and else block?

miles-grant-ibigroup (approved):
Working well for me, just one quick question

lib/editor/actions/map/stopStrategies.js
philip-cline: I had extracted it so that the `updateShapeDistTraveled` method would have the "correct" set of pattern stops when called. However, that method also looks up the stop using a filter on stopid, so splicing the stop out of there before hand doesn't actually matter. From a code perspective we could move the splice back after, but it may be more confusing inside of the `updateShapeDistTraveled` method since the patternStops will be "out of step" with the segments and controlPoints. I think as is might be easier to maintain.

PR #951, opened 2023-05-12
Don’t allow users to see all projects

miles-grant-ibigroup:
### Checklist

- [ ] Appropriate branch selected _(all PRs must first be merged to `dev` before they can be merged to `master`)_
- [ ] Any modified or new methods or classes have helpful JSDoc and code is thoroughly commented
- [ ] The description lists all applicable issues this PR seeks to resolve
- [ ] The description lists any configuration setting(s) that differ from the default settings
- [ ] e2e tests are all passing _(remove this if not merging to master)_

### Description

This PR resolves a bug where any user could see the full list of projects by visiting their "My Account" page. The code used is lifted from elsewhere.

I tried to make it a shared method, but the existing uses doing it via map, and this via filter made the shared implementation impractical.

It should also be noted this is not a complete solution -- the filtering really needs to happen server-side to avoid the data being leaked to the client to begin with

### DIFF
--- a/lib/public/components/UserAccount.js
+++ b/lib/public/components/UserAccount.js
@@ -68,6 +68,7 @@ export default class UserAccount extends Component<Props> {
 
   _onResetPassword = () => this.props.sendPasswordReset(this.props.user)
 
+  // eslint-disable-next-line complexity
   render () {
     const {accountTypes = {}, activeComponent, projectId, projects, user} = this.props
     const {profile} = user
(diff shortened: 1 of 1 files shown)


daniel-heppner-ibigroup (approved):
Looks good... I always forget that we don't have null coalescing on this project. Maybe add a bug to the server to fix this there?

philip-cline (approved):
LGTM, thanks for the quick fix

PR #986, opened 2023-09-07
Shared Stops Support File Upload

miles-grant-ibigroup:
### Checklist

- [x] e2e tests are all passing _(remove this if not merging to master)_

### Description

This PR expands upon the shared stops PR to add support for uploading a file to the textbook. Also adds support for uploading a file into the transformations form 🥳


lib/manager/components/ProjectSettingsForm.js
```diff
@@ -324,6 +330,26 @@ export default class ProjectSettingsForm extends Component<Props, State> {
...
+        <Panel>
+          <Panel.Heading><Panel.Title componentClass='h4'>{this.messages('fields.sharedStops.title')}</Panel.Title></Panel.Heading>
+          <ListGroup>
+            <ListGroupItem>
+              <FormGroup>
+                {/* TODO: on enter, textarea should NOT submit. This causes strange behavior when
+                editing in the textarea
+
+                see: https://github.com/ibi-group/datatools-ui/pull/977#discussion_r1288916749 */}
+                <CustomCSVForm
```
binh-dam-ibigroup, line 344: Is it me, or the placeholder text for the shared stops field is actually a template for a GTFS `stops.txt` file?

miles-grant-ibigroup: The component is shared with the transformations panel. It needs to show some form of csv. Any other ideas?

i18n/english.yml
```diff
@@ -96,6 +96,7 @@ components:
     title: Create a new snapshot
   CustomCSVForm:
     numLines: "%numLines% lines."
+    csvInvalid: CSV invalid!
```
philip-cline, line 99: We need this line added to the german and polish yamls too

lib/manager/components/ProjectSettingsForm.js
philip-cline: Why don't we just the pass a placeholder to the `CustomCSVForm` component so it varies based on context?

philip-cline: IMO it's pretty not-great to have the input showing the user sample text that would be invalid.

lib/manager/components/transform/CustomCSVForm.js
```diff
@@ -47,15 +63,29 @@ const CustomCSVForm = (props: Props) => {
           value={csvData} />
       </label>
       <div style={{marginBottom: '10px'}}>
-        <Button
+        <input
```
philip-cline, line 77: I was also more thinking of the file input (for transformations) as a separate input mode, not one that would populate the CSV text area, since one of the weaknesses of that text area is it can't display very large files (such as those for stop_times in a custom field preservation transformation). Can we make the file input a separate input mode for transformations and remove the population behaviour there?

philip-cline (changes requested):
A couple things to debate/change

lib/manager/components/transform/CustomCSVForm.js
miles-grant-ibigroup: Should we throw this to our weekly discussion call? I've tested this with large files and it is performant, and the feedback is helpful

philip-cline: Sure we can discuss on the call. I can send a feed that usually shows slow input into the csv text area as well

miles-grant-ibigroup:
Let me know what you think about this fix @philip-cline

philip-cline:
@miles-grant-ibigroup I'm still getting a page unresponsive error since we're loading the file directly which rly isn't great. We could do something or transformations like just pass the file to the backend and have it parse it to avoid that unresponsive error. However since this PR is focused on Shared Stops and the specific use case that would cause the lag in a transformation has a method which is fine rn, I'm fine with separating that into a later PR if you'd like

Lmk what you think @miles-grant-ibigroup

miles-grant-ibigroup:
Just to be clear, was this an issue before this PR?

philip-cline:
Good point, yes in a bit of a different way but yes

philip-cline (approved):
We can resolve the issues with large file csv uploads in a different PR

lib/manager/components/ProjectSettingsForm.js
```diff
@@ -324,6 +330,27 @@ export default class ProjectSettingsForm extends Component<Props, State> {
             </ListGroupItem>
           </ListGroup>
         </Panel>
+        <Panel>
+          <Panel.Heading><Panel.Title componentClass='h4'>{this.messages('fields.sharedStops.title')}</Panel.Title></Panel.Heading>
+          <ListGroup>
+            <ListGroupItem>
+              <FormGroup>
+                {/* TODO: on enter, textarea should NOT submit. This causes strange behavior when
```
binh-dam-ibigroup, line 340: This is because the `ProjectSettingsForm` has a keyboard listener (to emulate a missing `<form>` element?), but it doesn't look to me that that keyboard listener is needed.

lib/manager/components/ProjectSettingsForm.js
```diff
@@ -324,6 +330,27 @@ export default class ProjectSettingsForm extends Component<Props, State> {
...
+                editing in the textarea
+
+                see: https://github.com/ibi-group/datatools-ui/pull/977#discussion_r1288916749 */}
+                <CustomCSVForm
+                  name={'sharedStopsConfig'}
+                  csvData={model.sharedStopsConfig || ''}
```
binh-dam-ibigroup, line 345: Sort props

lib/manager/components/ProjectSettingsForm.js
```diff
@@ -324,6 +330,27 @@ export default class ProjectSettingsForm extends Component<Props, State> {
...
+                  hideSaveButton
+                  onChangeCsvData={(e) => this._onChangeTextInput(e)}
+                  onSaveCsvData={() => { alert('save') }}
```
binh-dam-ibigroup: What is this alert for?

lib/manager/components/ProjectSettingsForm.js
```diff
@@ -324,6 +330,27 @@ export default class ProjectSettingsForm extends Component<Props, State> {
...
+              <FormGroup>
+                {/* TODO: on enter, textarea should NOT submit. This causes strange behavior when
+                editing in the textarea
+
+                see: https://github.com/ibi-group/datatools-ui/pull/977#discussion_r1288916749 */}
+                <CustomCSVForm
+                  name={'sharedStopsConfig'}
+                  csvData={model.sharedStopsConfig || ''}
+                  hideSaveButton
+                  onChangeCsvData={(e) => this._onChangeTextInput(e)}
```
binh-dam-ibigroup:
```suggestion
                  onChangeCsvData={this._onChangeTextInput}
```

lib/manager/components/ProjectSettingsForm.js
```diff
@@ -182,8 +185,10 @@ export default class ProjectSettingsForm extends Component<Props, State> {
     this.setState(update(this.state, {model: {$merge: {defaultTimeZone}}}))
   }
 
-  _onChangeTextInput = ({target}: {target: HTMLInputElement}) => {
+  _onChangeTextInput = ({target}: {target: {name?: string, value: string}}) => {
```
binh-dam-ibigroup, line 190: I see why this is needed. Would it make sense to create a type such as `PartialInputEvent` ans reuse?

lib/manager/components/transform/CustomCSVForm.js
```diff
@@ -44,18 +71,33 @@ const CustomCSVForm = (props: Props) => {
...
             width: '400px'
           }}
-          value={csvData} />
+          value={csvData} />}
       </label>
       <div style={{marginBottom: '10px'}}>
-        <Button
+        <input
+          type='file'
+          name='file'
```
binh-dam-ibigroup, line 78: Sort props

lib/manager/components/transform/CustomCSVForm.js
```diff
@@ -44,18 +71,33 @@ const CustomCSVForm = (props: Props) => {
...
+          onChange={(e) => {
```
binh-dam-ibigroup, line 80: Use `useCallback`.

binh-dam-ibigroup (approved):
Looks good, see a few cleanup suggestions. We can discuss about a separate removal of the keyboard listener in `ProjectSettingsForm`.

lib/manager/components/ProjectSettingsForm.js
miles-grant-ibigroup: Let's discuss externally. Seems like an old issue

lib/manager/components/ProjectSettingsForm.js
miles-grant-ibigroup: Can go, good catch!

lib/manager/components/ProjectSettingsForm.js
miles-grant-ibigroup: Possibly! A great future refactor. I'll add a note

lib/manager/components/transform/CustomCSVForm.js
miles-grant-ibigroup: I tried this and got some unstable behavior. I'll add a note for now to investigate in the future

PR #1000, opened 2023-11-09
Remove duplicate key

AdrianaCeric:
### Checklist

### Description

I'm not sure how I originally found this. I'm also not sure if this has impact but it's probably best to remove?

### DIFF
--- a/lib/types/index.js
+++ b/lib/types/index.js
@@ -237,6 +237,7 @@ export type CustomFile = {
   useDuringServe: boolean
 }
 
+// TODO: Remove this eslint rule once Mastarm is upgraded/replaced
 /* eslint-disable no-use-before-define */
 export type Deployment = {
   buildGraphOnly?: boolean,
@@ -572,6 +573,8 @@ export type Label = {
   projectId: string,
   user?: string
 }
+
+// TODO: Remove this eslint rule once Mastarm is upgraded/replaced
 /* eslint-enable no-use-before-define */
 export type FeedInfo = {
   id: string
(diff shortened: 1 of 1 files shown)


lib/types/index.js
```diff
@@ -238,8 +237,6 @@ export type CustomFile = {
   useDuringServe: boolean
 }
 
-// TODO: Remove this eslint rule once https://github.com/babel/babel-eslint/pull/584
-// is merged in.
 /* eslint-disable no-use-before-define */
```
philip-cline, line 241: You should be able to remove this too since the PR above is merged

lib/types/index.js
```diff
@@ -575,9 +572,6 @@ export type Label = {
   projectId: string,
   user?: string
 }
-
-// TODO: Remove this eslint rule once https://github.com/babel/babel-eslint/pull/584
-// is merged in.
 /* eslint-enable no-use-before-define */
```
philip-cline, line 578: Likewise

philip-cline (approved):
Good catch, it's just flow typing so seems good.

lib/types/index.js
AdrianaCeric: I think we have an old version of flow since I still go the pre-commit lint error when I removed this

philip-cline: Ok we should change the TODO then to be updating eslint rather than waiting for the PR to be merged. I think the TODO is still relevant as long as the error is thrown

miles-grant-ibigroup (changes requested):
This is a really fitting 1000th pr for datatools. Agree with Phil on removing those comments until we upgrade/replace mastarm

AdrianaCeric:
woah happy 1000

miles-grant-ibigroup (approved):
It's beautiful! Happy 1000

philip-cline:
Happy 1000 everyone. Time to merge the documentation now

PR #1001, opened 2023-11-20
allow URL type for router and build config

miles-grant-ibigroup:
### Description

Allows us to use https://github.com/ibi-group/datatools-server/pull/577

Adds a third radio button to both configs. I tried to keep things abstract, but I think things are a bit messy. Looking for feedback on cleaning this up.

The issue is the way we determine if URL or text mode is active is by looking for the presence of URL data. Custom text should be preserved when URL is active.

### DIFF
--- a/lib/manager/components/deployment/CustomConfig.js
+++ b/lib/manager/components/deployment/CustomConfig.js
@@ -1,3 +1,4 @@
+/* eslint-disable complexity */
 // @flow
 
 import Icon from '@conveyal/woonerf/components/icon'
@@ -10,10 +11,10 @@ import {
   Radio
 } from 'react-bootstrap'
 import { LinkContainer } from 'react-router-bootstrap'
+import validator from 'validator'
 
 import * as deploymentActions from '../../actions/deployments'
 import {isValidJSONC} from '../../../common/util/json'
-
 import type {
   Deployment
 } from '../../../types'
@@ -30,6 +31,11 @@ const SAMPLE_ROUTER_CONFIG = `{
   }
 }`
 
+const CONFIG_OPTIONS = {
+  custom: 'custom',
+  url: 'url'
+}
+
 export default class CustomConfig extends Component<{
   deployment: Deployment,
   label: string,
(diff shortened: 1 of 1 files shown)


lib/manager/components/deployment/CustomConfig.js
```diff
@@ -65,39 +96,48 @@ export default class CustomConfig extends Component<{
...
+            checked={!useCustom && !useCustomUrl}
             name={name}
-            onChange={this._toggleCustomConfig}
+            onChange={(e) => this._toggleCustomConfig(e)}
             inline>
             Project default
           </Radio>
           <Radio
-            checked={useCustom}
+            checked={useCustom && !useCustomUrl}
```
philip-cline, line 116: Is the !useCustomUrl necessary here?

lib/manager/components/deployment/CustomConfig.js
```diff
@@ -38,24 +44,49 @@ export default class CustomConfig extends Component<{
...
+    if (name === 'customBuildConfig' && option === CONFIG_OPTIONS.custom) value = deployment[name] || SAMPLE_BUILD_CONFIG
+    if (name === 'customRouterConfig' && option === CONFIG_OPTIONS.custom) value = deployment[name] || SAMPLE_ROUTER_CONFIG
+
+    // If no option, clear everything
+    if (!option) {
+      updateDeployment(deployment, {[name + 'Url']: null, [name]: null})
+    }
+
+    // If custom content, clear URL
+    if (option === CONFIG_OPTIONS.custom) {
```
philip-cline, line 67: Would this be cleaner as an else if?

lib/manager/components/deployment/CustomConfig.js
```diff
@@ -38,24 +44,49 @@ export default class CustomConfig extends Component<{
...
+      updateDeployment(deployment, {[name + 'Url']: null, [name]: value})
+    }
+
+    // If custom URL, clear content
+    if (option === CONFIG_OPTIONS.url) {
+      updateDeployment(deployment, {[name + 'Url']: value})
```
philip-cline, line 73: Should we also clear `[name]: null ` here?

philip-cline (approved):
I tested this while I was testing the backend PR so thought I'd give a review. Looks good, left a couple of nit picky comments and questions. The naming convention of custom vs. customUrl can sometimes be confusing but I'm not sure it can be easily elucidated.

lib/manager/components/deployment/CustomConfig.js
miles-grant-ibigroup: Yes as both can be true in which case we want to show the URL field

lib/manager/components/deployment/CustomConfig.js
miles-grant-ibigroup: no

lib/manager/components/deployment/CustomConfig.js
miles-grant-ibigroup: No! This allows the custom content to be preserved if we switch back

daniel-heppner-ibigroup (approved):
It looks like it works for me thank you

PR #1026, opened 2025-02-26
Snap to Rail

josh-willis-arcadis:
### Checklist

- [ ] Appropriate branch selected _(all PRs must first be merged to `dev` before they can be merged to `master`)_
- [ ] Any modified or new methods or classes have helpful JSDoc and code is thoroughly commented
- [ ] The description lists all applicable issues this PR seeks to resolve
- [ ] The description lists any configuration setting(s) that differ from the default settings
- [ ] All tests and CI builds passing
- [ ] The description lists all relevant PRs included in this release _(remove this if not merging to master)_
- [ ] e2e tests are all passing _(remove this if not merging to master)_

### Description

Add support for a snap to rail feature when editing trip patterns. Uses OpenRailRouter, a fork of GraphHopper. Uses api that we host in EC2.

PR #1027, opened 2025-03-11
Snap To Rail

josh-willis-arcadis:
Add support for a snap to rail feature when editing trip patterns. Uses OpenRailRouter, a fork of GraphHopper.


lib/scenario-editor/utils/valhalla.js
```diff
@@ -198,26 +209,69 @@ export async function getSegment (
...
+          } else { // no URL or need default value so use openrailrouting test environment
+            graphHopperUrl = 'https://routing.openrailrouting.org/'
+          }
+        } else if (snapToOption === 'STREET') {
+          if (ROUTING_TYPE === 'STREET' && KEY && URL) {
+            graphHopperUrl = URL
+            graphHopperKey = KEY
+          } else { // no URL or need default value so use graphhopper test environment
+            graphHopperUrl = 'https://graphhopper.com/api/1/'
+            // include key here??? this will throw 401 i believe
```
miles-grant-ibigroup: You deleted the code that makes it required to have a graphhopper key! That's a required param

lib/scenario-editor/utils/valhalla.js
```diff
@@ -198,26 +209,69 @@ export async function getSegment (
  *
  * Example URL: https://graphhopper.com/api/1/route?point=49.932707,11.588051&point=50.3404,11.64705&vehicle=car&debug=true&&type=json
  */
-export function routeWithGraphHopper (points: Array<LatLng>, avoidMotorways?: boolean): ?Promise<GraphHopperResponse> {
+export function routeWithGraphHopper (
+  points: Array<LatLng>,
+  avoidMotorways?: boolean,
+  snapToOption: string
```
miles-grant-ibigroup: I'm not sure you're allowed to put an optional before a required parameter

lib/scenario-editor/utils/valhalla.js
```diff
@@ -198,26 +209,69 @@ export async function getSegment (
...
+  let graphHopperKey = ''
+
   // Use custom url if it exists, otherwise default to the hosted service.
-  const GRAPH_HOPPER_URL = process.env.GRAPH_HOPPER_URL || 'https://graphhopper.com/api/1/'
+  if (process.env.GRAPH_HOPPER_ALTERNATES) {
+    // $FlowFixMe This is a bit of a hack and not how env variables are supposed to work, but the yaml loader supports it.
+    const alternates: Array<GraphHopperAlternateServer> = process.env.GRAPH_HOPPER_ALTERNATES
+    alternates.forEach(alternative => {
+      const { BBOX, ROUTING_TYPE, URL, KEY } = alternative
+      if (BBOX.length !== 4) {
```
miles-grant-ibigroup: The edits to Valhalla are quite clunky. Couldn't we instead mostly keep things as is and only tweak the url/key after the existing block?

lib/scenario-editor/utils/valhalla.js
```diff
@@ -237,7 +291,19 @@ export function routeWithGraphHopper (points: Array<LatLng>, avoidMotorways?: bo
...
+  return points.every(
+    (point) => !coordIsOutOfBounds(
+      point,
+      L.latLngBounds(
+        [BBOX[1], BBOX[0]],
+        [BBOX[3], BBOX[2]]
+      )
+    )
+  )
+}
```
miles-grant-ibigroup: Why was this imported into this file?

lib/editor/actions/map/index.js
```diff
@@ -162,17 +162,17 @@ export function handleControlPointDrag (
   patternCoordinates: any
 ) {
   return function (dispatch: dispatchFn, getState: getStateFn) {
-    const {avoidMotorways, currentDragId, followStreets} = getState().editor.editSettings.present
+    const {avoidMotorways, currentDragId, snapToOption} = getState().editor.editSettings.present
```
miles-grant-ibigroup: Could we call this `followOption` instead of `snapToOption`?

lib/editor/actions/map/stopStrategies.js
```diff
@@ -260,7 +261,7 @@ export function addStopToPattern (pattern: Pattern, stop: GtfsStop, index?: ?num
       } else {
         dispatch(updatePatternStops(pattern, patternStops))
         // Otherwise, check if a shape ought to be created. Then, save.
-        if (patternStops.length === 2 && followStreets) {
+        if (patternStops.length === 2 && snapToOption) {
```
miles-grant-ibigroup: This is anticipating `followStreets` to be a boolean. To reduce accidental regressions maybe we leave `followStreets` as is and create a new variable `followRail`? I know it's clunkier but it lets us touch much less code

lib/editor/util/index.js
```diff
@@ -19,6 +19,11 @@ export const CLICK_OPTIONS: Array<string> = [
   'ADD_STOPS_AT_INTERVAL',
   'ADD_STOPS_AT_INTERSECTIONS'
 ]
+export const SNAP_TO_OPTIONS: Array<string> = [
```
miles-grant-ibigroup: Why is this an array instead of an enum?

lib/editor/components/pattern/EditSettings.js
```diff
@@ -75,6 +74,21 @@ export default class EditSettings extends Component<Props, State> {
...
+        <FormControl
+          componentClass='select'
+          value={snapToOption}
+          name={'snapToOption'}
+          onChange={this._onSelectChange}>
+          {SNAP_TO_OPTIONS.map(v => {
+            return (
+              <option
+                key={v}
+                value={v}>{toSentenceCase(v.replace(/_/g, ' '))}
```
miles-grant-ibigroup: You could avoid this by using an enum and then having a map from enum to string

miles-grant-ibigroup: with JS you could even have one large object that has both enum value and string!

lib/editor/components/pattern/EditSettings.js
```diff
@@ -83,7 +97,7 @@ export default class EditSettings extends Component<Props, State> {
             // this state would cause the entire shape to disappear).
             disabled={
               (s.type === 'hideInactiveSegments' && noSegmentIsActive) ||
-              (s.type === 'avoidMotorways' && !followStreets)
+              (s.type === 'avoidMotorways' && snapToOption !== 'STREET')
```
miles-grant-ibigroup: No magic strings! Let's use enums instead if we stick with `snapToOption`

lib/editor/components/pattern/EditShapePanel.js
```diff
@@ -104,7 +104,7 @@ export default class EditShapePanel extends Component<Props> {
           return {lng: stop.stop_lon, lat: stop.stop_lat}
         })
       : []
-    this.drawPatternFromStops(activePattern, stopLocations, editSettings.present.followStreets)
+    this.drawPatternFromStops(activePattern, stopLocations)
```
miles-grant-ibigroup: Uh oh! I'm not sure `followStreets` was meant to be removed. Maybe let's keep `followStreets` as is and add a new boolean for `followRail`. It makes the UI work a bit more tricky but it helps us avoid touching too many files

lib/editor/reducers/settings.js
```diff
@@ -96,6 +96,11 @@ export const reducers = {
     action: ActionType<typeof updateEditSetting>
   ): EditSettingsState {
     const {setting, value} = action.payload
-    return update(state, { [setting]: {$set: value} })
+    // RAIL-TODO: make sure followStreet is properly set.
+    // RAIL-TODO: use in a lot of places, make sure nothing breaks????
+    return update(state, {
+      [setting]: {$set: value},
+      followStreets: {$set: value === 'STREET'}
+    })
```
miles-grant-ibigroup: Yes let's definitely leave `followStreets` alone

miles-grant-ibigroup (changes requested):
I see quite a bit of potential to reduce the number of files touched. I'd start by leaving `followStreets` untouched. That will also make the edits to `valhalla` less dangerous. We can leave that file mostly untouched and merely add a short check after the alternates section to switch to rail if `followRail` is true. 

That way we can also configure rail routing within alternates. Just have to add a rail flag on each graphhopper alternate!

lib/scenario-editor/utils/valhalla.js
josh-willis-arcadis: the rail routing server doesn't require a key. only the street routing server.

lib/scenario-editor/utils/valhalla.js
josh-willis-arcadis: this is a function from the dev branch that is not in the mtc branch. the bones of this logic is from the dev branch.

lib/scenario-editor/utils/valhalla.js
josh-willis-arcadis: makes sense. will flow throw an error here?

lib/scenario-editor/utils/valhalla.js
josh-willis-arcadis: I agree. A lot of if else. I can try a different approach. I copied code from the dev branch (because this branch didn't support `GRAPH_HOPPER_ALTERNATES`) and started from there.

lib/editor/util/index.js
josh-willis-arcadis: Copied the format from the `CLICK_OPTIONS` array from above to display the values in the dropdown. enum could work here as well.

lib/editor/components/pattern/EditShapePanel.js
josh-willis-arcadis: think I got too excited with my idea for `snapToOptions`. I agree that we are touching a lot of files.

lib/editor/actions/map/stopStrategies.js
josh-willis-arcadis: I started off `followRail` and swapped to `snapToOption` trying to make it easier to add different options later, but removing `followStreets` is not the best option here.

josh-willis-arcadis: I'll get the new var `followRail` back in here

lib/editor/actions/map/index.js
josh-willis-arcadis: great idea

lib/editor/components/pattern/EditSettings.js
```diff
@@ -75,6 +79,25 @@ export default class EditSettings extends Component<Props, State> {
     const noSegmentIsActive = !patternSegment && patternSegment !== 0
     return (
       <div>
+        <ControlLabel><small>Snap to options</small></ControlLabel>
```
miles-grant-ibigroup: Nitpick: should we shorten this string?
```suggestion
        <ControlLabel><small>Snap to</small></ControlLabel>
```

lib/editor/components/pattern/EditSettings.js
```diff
@@ -75,6 +79,25 @@ export default class EditSettings extends Component<Props, State> {
...
+        <FormControl
+          componentClass='select'
+          value={currentSelection}
+          name={'followOption'}
```
miles-grant-ibigroup: Nitpick: cleanup
```suggestion
          name='followOption'
```

lib/editor/components/pattern/EditSettings.js
```diff
@@ -75,6 +79,25 @@ export default class EditSettings extends Component<Props, State> {
...
+          onChange={(evt) => {
+            const value = evt.target.value
```
miles-grant-ibigroup: nitpick: you can extract `value` here

lib/editor/util/map.js
```diff
@@ -284,6 +291,7 @@ export async function recalculateShape ({
   dragId?: ?string,
   updatedControlPoints: ?Array<ControlPoint>
 }> {
+  console.log('update shape!!!!!!!!!')
```
miles-grant-ibigroup: Oops!

lib/editor/util/map.js
```diff
@@ -389,12 +397,14 @@ export async function recalculateShape ({
   } else {
     throw new Error('Invalid editType. Must be either "update" or "delete".')
   }
+  console.log('arre we goign to valhalla???')
```
miles-grant-ibigroup: Ooops!

lib/scenario-editor/utils/valhalla.js
```diff
@@ -198,26 +210,73 @@ export async function getSegment (
...
-  if (!process.env.GRAPH_HOPPER_KEY) {
-    throw new Error('GRAPH_HOPPER_KEY not set')
-  }
+
   // Use custom url if it exists, otherwise default to the hosted service.
-  const GRAPH_HOPPER_URL = process.env.GRAPH_HOPPER_URL || 'https://graphhopper.com/api/1/'
+  let graphHopperUrl = process.env.GRAPH_HOPPER_URL || 'https://graphhopper.com/api/1/'
+  let graphHopperKey = process.env.GRAPH_HOPPER_KEY
+
+  console.log('graphHopperUrl::::::', graphHopperUrl)
```
miles-grant-ibigroup: Oops!

lib/scenario-editor/utils/valhalla.js
```diff
@@ -198,26 +210,73 @@ export async function getSegment (
...
+          graphHopperUrl = alternative.URL
+        }
+        if (alternative.KEY) {
+          graphHopperKey = alternative.KEY
+        }
+      }
+    })
+  }
+
+  console.log('graphHopperUrl::::::', graphHopperUrl)
```
miles-grant-ibigroup: Oops!

miles-grant-ibigroup (approved):
I'm noticing that the default OpenRailServer server is missing, but maybe that's ok? A few small nitpicks and cleanups but once those are resolved this is good to go!

lib/common/util/text.js
binh-dam-ibigroup, line 1: If you want to use this file instead of the existing one, delete the original file `to-sentence-case` and migrate all imports (that touches quite a few files). Alternately, use the existing file and delete this one.

lib/editor/actions/map/stopStrategies.js
```diff
@@ -336,6 +336,7 @@ export function addStopToPattern (pattern: Pattern, stop: GtfsStop, index?: ?num
             defaultToStraightLine: false,
             editType: 'update',
             followStreets,
+            followRail: editSettings.present.followRail,
```
binh-dam-ibigroup: Extract `followRail` with the deconstruct at line 225 above (where `followStreets` is also extracted) (2 instances).

lib/editor/components/pattern/EditSettings.js
```diff
@@ -75,6 +79,24 @@ export default class EditSettings extends Component<Props, State> {
...
       <div>
+        <ControlLabel><small>Snap to</small></ControlLabel>
+        <FormControl
+          componentClass='select'
+          value={currentSelection}
+          name='followOption'
+          onChange={(evt) => {
+            // Update both settings based on selection
+            updateEditSetting({setting: 'followStreets', value: evt.target.value === SNAP_TO_OPTIONS.STREET})
+            updateEditSetting({setting: 'followRail', value: evt.target.value === SNAP_TO_OPTIONS.RAIL})
```
binh-dam-ibigroup: Extract a class method to handle the event.

lib/scenario-editor/utils/valhalla.js
```diff
@@ -198,26 +210,69 @@ export async function getSegment (
...
+      // Check if points are within bounds and routing type matches
+      if (
+        points.every(
+          (point) =>
+            !coordIsOutOfBounds(
+              point,
+              L.latLngBounds(
+                [alternative.BBOX[1], alternative.BBOX[0]],
+                [alternative.BBOX[3], alternative.BBOX[2]]
+              )
```
binh-dam-ibigroup: Extract a variable, so that the bounds are not re-created for every `point`. After that, you might be able to shrink the function call to just one line or two.

lib/editor/actions/map/stopStrategies.js
```diff
@@ -377,6 +378,7 @@ export function addStopToPattern (pattern: Pattern, stop: GtfsStop, index?: ?num
...
```
binh-dam-ibigroup: I think you are missing a change in this file at line 275, otherwise the `followRail` option appears unused. See what you did in #1026 at https://github.com/ibi-group/datatools-ui/blob/f3315efe839100c2c785283ca17ccb58f8131515/lib/editor/actions/map/stopStrategies.js#L281.

binh-dam-ibigroup (changes requested):
I think you are missing a change to fetch a polyline with the `followRail` option (see comment). See also other suggested refactors.

lib/editor/actions/map/stopStrategies.js
josh-willis-arcadis: great catch. addressed in [39722d9](https://github.com/ibi-group/datatools-ui/pull/1027/commits/39722d9836c7dbc9c6a6c00349f5fafa60c8c4f1)

lib/editor/components/pattern/EditSettings.js
```diff
@@ -75,6 +85,20 @@ export default class EditSettings extends Component<Props, State> {
     const noSegmentIsActive = !patternSegment && patternSegment !== 0
     return (
       <div>
+        <ControlLabel><small>Snap to</small></ControlLabel>
+        <FormControl
+          componentClass='select'
+          value={currentSelection}
+          name='followOption'
+          onChange={this._onFollowOptionChange}>
```
binh-dam-ibigroup: Nit - Sort props and put closing `>` on new line.

binh-dam-ibigroup (approved):
Much better now, and all modes work. Thanks for the all the work.

binh-dam-ibigroup (changes requested):
Apologies, one requirement we overlooked. Could you "enable “Snap To Rail” function when editing routes with rail GTFS route_type (0 - Light Rail/Streetcar, 1 - Subway/Metro , 2 - Rail) and not enable it for all other route_type values", please. Thank you!

lib/editor/components/pattern/EditSettings.js
```diff
@@ -62,7 +64,7 @@ export default class EditSettings extends Component<Props, State> {
 
   render () {
     // Edit Settings passed in are present
-    const {editSettings, patternSegment, updateEditSetting} = this.props
+    const {editSettings, patternSegment, updateEditSetting, activeEntity} = this.props
```
binh-dam-ibigroup: Nit - sort props

binh-dam-ibigroup (approved):
Nice

PR #948, opened 2023-05-03
Update valhalla.js Graphhopper to V7

mvanlaar:
### Checklist

- [ ] Appropriate branch selected _(all PRs must first be merged to `dev` before they can be merged to `master`)_
- [ ] Any modified or new methods or classes have helpful JSDoc and code is thoroughly commented
- [ ] The description lists all applicable issues this PR seeks to resolve
- [ ] The description lists any configuration setting(s) that differ from the default settings
- [ ] All tests and CI builds passing
- [ ] The description lists all relevant PRs included in this release _(remove this if not merging to master)_
- [ ] e2e tests are all passing _(remove this if not merging to master)_

### Description

Version 7 of graphhopper api no longer supported vehicle, you have to use profile paramter instead. 

See: https://github.com/graphhopper/graphhopper/blob/master/CHANGELOG.md

PR #1043, opened 2025-10-04
fix(graphhopper): Rename vehicle param to profile for routing requests with Graphhopper

marecabo:
### Checklist

- [ ] All tests and CI builds passing [someone needs to trigger them]

### Description

This fixes Street Snapping when editing shapes in the GTFS editor.

When I tried it with a local Graphhopper instance before, I got the error message "bad request: profile parameter required" from the routing request.

At some point, Graphhopper switched from a `vehicle` parameter to `profile`, see https://github.com/graphhopper/graphhopper/pull/1934 and https://github.com/graphhopper/graphhopper/blob/master/CHANGELOG.md#70-14-mar-2023

There is already a PR https://github.com/ibi-group/datatools-ui/pull/948 for updating this code, but it has not been edited since over two years. This edit aims to be minimally invasive to restore the routing functionality.


miles-grant-ibigroup:
Closes #948

miles-grant-ibigroup (approved):
Thank you so much for this work!

binh-dam-ibigroup (approved):
Nice fix, thank you

lib/scenario-editor/utils/valhalla.js
```diff
@@ -257,7 +257,7 @@ export function routeWithGraphHopper (points: Array<LatLng>, avoidMotorways?: bo
 
   const params = {
     key: graphHopperKey,
-    vehicle: 'car',
+    profile: 'car',
```
binh-dam-ibigroup, line 259: @marecabo You might need to merge with our `dev` branch, or if you are not able to do it, we'll have to do a separate PR (#1053). Merging your PR as is will fail with a "duplicate key" error:
```js
  const params = {
    key: graphHopperKey,
    profile: 'car',
    debug: true,
    type: 'json',
    profile: '' // <== Bad merge!
  }
```

binh-dam-ibigroup (changes requested):
Merging as is will result in a bad merge, see my comment.

lib/scenario-editor/utils/valhalla.js
binh-dam-ibigroup: Apologies for the roughness... It was a good call to run the tests haha!

marecabo: Thanks, I see that makes no sense, having the same key twice. If I just need to remove it, I will look at it tomorrow.

marecabo: @binh-dam-ibigroup Can I trigger the CI now myself, when I push new commits?

marecabo: @binh-dam-ibigroup It should be fixed now, but I cannot trigger the CI to test it.

marecabo: @binh-dam-ibigroup And just fixed the dangling comma as well.

miles-grant-ibigroup:
The failures we're seeing now are normal! Since this PR is external. Once @binh-dam-ibigroup approves this is good to merge. Thanks so much for all your work!

binh-dam-ibigroup (approved):
Thanks for the changes! (Tests succeeded on #1053)

PR #1057, opened 2026-05-06
Repair Interpolation

miles-grant-ibigroup:
Allows for use of https://github.com/ibi-group/datatools-server/pull/647


lib/editor/components/pattern/NormalizeStopTimesModal.js
```diff
@@ -16,36 +16,42 @@ type Props = {
...
 
-type State = { interpolateStopTimes: boolean, patternStopIndex: number, show: boolean }
+type State = { ignoreNonBlank: boolean, interpolateStopTimes: boolean, patternStopIndex: number, show: boolean }
 
 export default class NormalizeStopTimesModal extends Component<Props, State> {
   messages = getComponentMessages('NormalizeStopTimesModal')
 
   state = {
     interpolateStopTimes: false,
+    ignoreNonBlank: false,
```
josh-willis-arcadis, line 25: nit sort.

josh-willis-arcadis (approved):
Code looks good. 1 nit is all.

i18n/english.yml
```diff
@@ -804,7 +804,7 @@ components:
       substitutionInvalid: Some substitution patterns are invalid.
   NormalizeStopTimesModal:
     close: Close
-    interpolateStopTimes: Interpolate stop times between timepoints?
+    interpolateStopTimes: Assume timepoint travel times are since last timepoint? (Interpolate non-timepoint stops)
```
binh-dam-ibigroup: IMO the new wording is not needed and introduces more confusion instead.

The UI currently says that travel time is from the previous stop in a pattern, however "travel times are since last timepoint" could imply that the travel time from the first stop, second stop, etc. after a timepoint are relative to that. As implemented, that is not the case, as all travel times are added, so that the next timepoint time is the sum of travel times from the previous timepoint.

i18n/english.yml
```diff
@@ -820,7 +820,7 @@ components:
       based on the shape distance and the default travel times. This speed is
       then applied to the shape distance traveled for each intermediate non-timepoint
       stop to provide interpolated travel times. The default travel time for non-timepoint
-      stops will not be modified.
+      stops will not be modified. This feature assumes that the timepoint travel times refer to the time since the last timepoint!
```
binh-dam-ibigroup: Ditto here.

i18n/english.yml
```diff
@@ -820,7 +820,7 @@ components:
       based on the shape distance and the default travel times. This speed is
       then applied to the shape distance traveled for each intermediate non-timepoint
       stop to provide interpolated travel times. The default travel time for non-timepoint
```
binh-dam-ibigroup: We probably need to clarify what "default travel time" means or change the wording to something like 

"...the implicit speed between timepoints based on the shape distance and travel times entered for each pattern stop."

"The travel time that is entered/specified for non-timepoint stops will not be modified."

lib/editor/components/pattern/NormalizeStopTimesModal.js
```diff
@@ -89,12 +95,21 @@ export default class NormalizeStopTimesModal extends Component<Props, State> {
...
-            <span style={{paddingBottom: '2px'}}>{this.messages('interpolateStopTimes')}</span>
+            <label htmlFor='interpolate' style={{fontWeight: 'normal', paddingBottom: '2px'}}>{this.messages('interpolateStopTimes')}</label>
+          </div>
+          <div style={{alignContent: 'center', alignItems: 'center', display: 'flex', marginTop: -20, marginBottom: -20}}>
+            <Checkbox
+              id='ignoreNonBlank'
+              onChange={this._onChangeNonBlank}
+              value={this.state.ignoreNonBlank}
+            />
+            <label htmlFor='ignoreNonBlank' style={{fontWeight: 'normal', paddingBottom: '2px'}}>Ignore non-blank stop-times?</label>
```
binh-dam-ibigroup: Double negative. Maybe change "Ignore non-blank stop times" to "Preserve existing stop times outside of timepoints"? Move the text to i18n file.

binh-dam-ibigroup (changes requested):
Wording change for the interpolate checkbox is not needed IMO. Rephrase the "ignore non-blank" text, and move it to i18n.

i18n/english.yml
miles-grant-ibigroup: I can make it clearer I'm referring to pattern times and not stop times. I think it's an important distinction to make, as this differs from previous behavior. This mode will ignore all non-timepoint pattern times. To make that work all timepoint pattern data must be relative to the last timepoint. How's this?

```suggestion
    interpolateStopTimes: Assume timepoint travel times are relative to the previous timepoint? (This willl generate stop times for non-timepoint stops)
```

i18n/english.yml
miles-grant-ibigroup: I think it's an important thing to mention? Because if the user sets their timepoint travel times to be relative to the last non-timepoint stop, the results will be awful

i18n/english.yml
miles-grant-ibigroup: No travel time is adjusted at all. Only stop times are updated

lib/editor/components/pattern/NormalizeStopTimesModal.js
miles-grant-ibigroup: This is not what this does though. They are 2 separate modes. I will clarify

i18n/english.yml
binh-dam-ibigroup: ~It is not obvious IMO. The user docs suggest that the Travel Time field is from the previous stop in the sequence: https://github.com/ibi-group/datatools-ui/blob/7495634765d4fee5bcd3f269ebc617f84cb018fe/docs/user/editor/patterns.md?plain=1#L101. If this interpretation is changed with this PR, update the documentation and the UI label (and other relevant code) so that there is no ambiguity.~

i18n/english.yml
binh-dam-ibigroup: How about this?
```suggestion
    interpolateStopTimes: Interpolate stop times between timepoints (Ignore all travel/dwell times except for timepoints)
```

i18n/english.yml
```diff
@@ -817,10 +818,10 @@ components:
       you can normalize the stop times to bring them into alignment with the
       updated travel times reflected in the pattern stops.
     usageExplanationTwo: Interpolating stop times calculates the implicit speed between timepoints
-      based on the shape distance and the default travel times. This speed is
+      based on the shape distance and the default travel times of timepoints. This speed is
       then applied to the shape distance traveled for each intermediate non-timepoint
-      stop to provide interpolated travel times. The default travel time for non-timepoint
-      stops will not be modified. This feature assumes that the timepoint travel times refer to the time since the last timepoint!
+      stop to provide interpolated stop times. 
+      This feature assumes that the timepoint travel times refer to the time since the last timepoint!
```
binh-dam-ibigroup:
```suggestion
      This feature ignores all travel/dwell times except for timepoints!
```

i18n/english.yml
binh-dam-ibigroup: It is probably worth updating the documentation at https://github.com/ibi-group/datatools-ui/blob/7495634765d4fee5bcd3f269ebc617f84cb018fe/docs/user/editor/patterns.md?plain=1#L101 to reflect this difference in behavior.

i18n/english.yml
```diff
@@ -804,7 +804,8 @@ components:
       substitutionInvalid: Some substitution patterns are invalid.
   NormalizeStopTimesModal:
     close: Close
-    interpolateStopTimes: Interpolate stop times between timepoints?
+    ignoreNonBlank: Only update blank stop times
```
binh-dam-ibigroup: Consider renaming `ignoreNonBlank` => `updateBlankOnly`.

binh-dam-ibigroup (approved):
See comments for further improving the text. Consider updating the corresponding paragraph in the user documentation too.

